PR #423, opened 2026-04-19
fix(about): inject git hash via generated module, not Vite define (#422)

### LINKED ISSUE #422: fix(about): Vite define for __GIT_HASH__ and __SENTRY_DSN__ not replaced in packaged builds
## Symptom

About dialog (Prose → About Prose) shows `Version 1.0.0 (dev)` in packaged TestFlight builds, instead of the expected git hash. Confirmed against Build 18 on TestFlight.

## Root cause

`electron-vite.config.ts:57-63` uses Vite's \`renderer.define\` to inject `__GIT_HASH__` and `__SENTRY_DSN__` at build time:

\`\`\`ts
renderer: {
  ...
  define: {
    __GIT_HASH__: JSON.stringify(getGitHash()),
    __SENTRY_DSN__: JSON.stringify(process.env.SENTRY_DSN || '...')
  }
}
\`\`\`

But the replacement is **not taking effect**. Verified in the shipped bundle \`out/renderer/assets/index-DiWT6UV0.js\`:

- Line 131193: `const gitHash = typeof __GIT_HASH__ !== "undefined" ? __GIT_HASH__ : "dev";` — identifier never replaced
- Line 7137: `const SENTRY_DSN = typeof __SENTRY_DSN__ !== "undefined" ? __SENTRY_DSN__ : "<hardcoded fallback>";` — same pattern

Both fall through to their hardcoded fallbacks at runtime. For Sentry this is benign (the fallback IS the correct project DSN). For the About dialog it surfaces the visible `"(dev)"` text in production.

## Scope

- \`src/renderer/components/settings/AboutDialog.tsx:13\` (git hash display)
- \`src/renderer/lib/sentry.ts:9-12\` (DSN fallback)
- \`electron-vite.config.ts:57-63\` (broken define)

## Proposed fix

Switch both injections to Vite's standard \`import.meta.env.VITE_*\` mechanism, which is well-documented and reliable across electron-vite versions:

\`\`\`ts
// electron-vite.config.ts
process.env.VITE_GIT_HASH = getGitHash()
process.env.VITE_SENTRY_DSN = process.env.SENTRY_DSN || '<default-dsn>'
\`\`\`

Then consume via \`import.meta.env.VITE_GIT_HASH\` / \`.VITE_SENTRY_DSN\` in the renderer. Remove the broken \`renderer.define\` block.

## Verification

After fix, rebuild with \`npm run build:mas\` and confirm the About dialog shows \`Version 1.0.0 (<7-char-hash>)\`, not \`(dev)\`.

mrangelmarino:
Closes #422

## Problem

About dialog shows \`Version 1.0.0 (dev)\` in packaged Build 18 (TestFlight), instead of the git hash. Confirmed by the user in TestFlight QA.

## Root cause

\`electron-vite.config.ts\` used Vite's \`renderer.define\` to inject \`__GIT_HASH__\` and \`__SENTRY_DSN__\` tokens. But the tokens weren't being replaced in the shipped bundle — confirmed in \`out/renderer/assets/index-DiWT6UV0.js\` (Build 18):

- Line 131193: \`const gitHash = typeof __GIT_HASH__ !== "undefined" ? __GIT_HASH__ : "dev";\` — identifier never replaced
- Line 7137: \`const SENTRY_DSN = typeof __SENTRY_DSN__ !== "undefined" ? __SENTRY_DSN__ : "<hardcoded>";\` — same pattern

Both fell through to their hardcoded fallbacks. Sentry's fallback IS the correct production DSN so that was benign; About surfaced the visible \`"(dev)"\`.

I didn't diagnose WHY Vite's define wasn't firing (electron-vite v5 specifics, config resolution timing, or something else) — but switching away from define entirely is both simpler and guaranteed-correct.

## Fix

Build-time generated TypeScript module, mirroring the existing \`src/main/env.ts\` pattern for \`IS_MAS_BUILD\`:

- New \`build:buildinfo\` script writes \`src/renderer/buildInfo.generated.ts\` with the short git hash, runs before \`electron-vite build\` and \`electron-vite dev\`
- \`AboutDialog\` imports \`GIT_HASH\` from the generated module (no more \`declare const\` or \`|| 'dev'\` fallback)
- Removed \`renderer.define\` from \`electron-vite.config.ts\` and associated type declarations
- Sentry renderer uses the hardcoded production DSN directly (removes broken define indirection; behavior unchanged)
- Gitignore \`src/renderer/buildInfo.generated.ts\`
- Bump MAS \`buildVersion\` 18 → 19

## Verification

Post-build:

\`\`\`
$ cat src/renderer/buildInfo.generated.ts
// Auto-generated — do not edit
export const GIT_HASH = "45c7845"

$ grep 'const GIT_HASH' out/renderer/assets/*.js
const GIT_HASH = "45c7845";
\`\`\`

Confirmed the hash is actually inlined into the bundle — the previous define-based approach left \`__GIT_HASH__\` as a dangling undefined identifier.

## Test plan

- [ ] Local dev server (\`npm run dev\`) — About dialog shows \`Version 1.0.0 (<hash>)\`
- [ ] Packaged TestFlight Build 19 — About dialog shows \`Version 1.0.0 (<hash>)\`, not \`(dev)\`


mrangelmarino:
/review 

PR #428, opened 2026-04-19
fix(mcp): resolve full node path for Claude Desktop install

### LINKED ISSUE #427: fix(mcp): resolve full node path when installing MCP for Claude Desktop
## Problem

Version managers like fnm/nvm put node outside the default PATH that Claude Desktop searches. When Prose writes the `claude_desktop_config.json` entry for its MCP stdio bridge, an unqualified `node` reference fails to launch on machines where Claude Desktop can't find node on its own PATH.

## Fix

Resolve the full path via `which node` at install time and write the absolute path into `claude_desktop_config.json` instead of bare `node`.

Also: add `build:mcp-stdio` to the `dev` script so the MCP stdio bridge is always available after a clean checkout.

## Origin

This change was originally drafted as commit \`694a714\` on the \`issue-369-remarkable-early-access\` branch but was orthogonal to the reMarkable feature work. Splitting it out so it can ship on its own merits without coupling to the reMarkable rollup.

mrangelmarino:
## Summary

Resolve the absolute path of `node` (via `which node`) when writing Prose's entry into `claude_desktop_config.json`. Without this, version-managed node installs (fnm, nvm) aren't on Claude Desktop's default PATH and the MCP server fails to launch.

Also: add `build:mcp-stdio` to the `dev` script so the stdio bridge is always built after a clean checkout.

## Origin

This was originally drafted as commit `694a714` on `issue-369-remarkable-early-access` but was orthogonal to that branch's reMarkable feature work. Cherry-picked to its own branch so it can ship on its own merits without coupling to the reMarkable rollup. The original commit is dropped from `issue-369-remarkable-early-access` in a follow-up step.

Fixes #427

## Test plan

- [ ] Build succeeds (verified locally)
- [ ] Install MCP from Settings → Integrations → Claude Desktop with fnm/nvm-managed node; confirm `claude_desktop_config.json` contains the full node path, not bare `node`
- [ ] Restart Claude Desktop; confirm Prose MCP server connects

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "type": "module",
   "main": "./out/main/index.js",
   "scripts": {
-    "dev": "npm run build:env && npm run build:buildinfo && npm run build:preload && electron-vite dev",
+    "dev": "npm run build:env && npm run build:buildinfo && npm run build:mcp-stdio && npm run build:preload && electron-vite dev",
     "dev:kill": "kill $(cat .dev.pid 2>/dev/null) 2>/dev/null || true; lsof -ti:9222 | xargs kill -9 2>/dev/null || true; lsof -ti:5173 | xargs kill -9 2>/dev/null || true",
     "dev:restart": "npm run dev:kill && sleep 2 && npm run dev",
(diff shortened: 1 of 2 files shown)


mrangelmarino:
/review 

PR #480, opened 2026-05-12
fix(editor): stabilize frontmatter + restore comment modal click (#449, #456)

mrangelmarino:
## Summary

- **#449 — Frontmatter polish + layout shift bug**: Introduced a `stableFrontmatter` state that only updates on document ID changes, preventing the `FrontmatterEditor` from receiving new prop references during body edits. Changed `FrontmatterEditor`'s `key` from `document.path` to `document.documentId` for more stable mounting. Also aligned the collapsed/read-only frontmatter views with the expanded edit-mode (fixed-width `w-32` key column + `flex-1` value column).
- **#456 — Comments click regression**: Added a ProseMirror plugin to the Comment extension that fires a custom `editor:comment-click` event when a `.comment-mark` span is clicked. `Editor.tsx` listens for this event and opens `AddCommentDialog` in a new view-mode, showing the existing comment text and offering a "Remove Comment" button.

## Changes

**`src/renderer/components/editor/Editor.tsx`**
- `stableFrontmatter` state + `stableFrontmatterDocIdRef`: frontmatter prop to `FrontmatterEditor` only updates on document ID change
- `FrontmatterEditor` key changed from `document.path || 'new'` to `document.documentId`
- `showFrontmatter` derived from `stableFrontmatter` instead of `document.frontmatter`
- `handleFrontmatterSave` updates `stableFrontmatter` alongside the store
- New `viewingCommentId` / `viewingCommentText` state for view-mode dialog
- `editor:comment-click` event listener opens dialog in view mode
- `openAddCommentDialog` clears view-mode state before opening add mode

**`src/renderer/extensions/comments/extension.ts`**
- Added `addProseMirrorPlugins` returning a plugin with `handleDOMEvents.click` that detects `.comment-mark` clicks and dispatches `editor:comment-click` CustomEvent

**`src/renderer/components/editor/AddCommentDialog.tsx`**
- New `existingCommentId` + `existingCommentText` props for view mode
- Read-only textarea in view mode; "Remove Comment" + "Close" buttons (replaces add-mode "Cancel"/"Add Comment")
- `handleRemove` calls `editor.commands.unsetComment(id)`

**`src/renderer/components/editor/FrontmatterEditor.tsx`**
- Collapsed view rows: `w-32 truncate` key + `:` separator + `flex-1 truncate` value (matches edit-mode alignment)

**`src/renderer/components/editor/FrontmatterDisplay.tsx`**
- Same alignment fix as FrontmatterEditor collapsed view

## Test plan

[...]

PR #479, opened 2026-05-12
feat(mcp): nest containers, fix suggest_edit, surface comments (#458, #448, #457, #455)

mrangelmarino:
## Summary

Four MCP tool layer fixes, one per commit:

- **#458 — read_document nesting**: Container nodes (blockquote, bulletList, orderedList, listItem, taskItem) now carry a `children[]` array in the `read_document` response. Descendants are no longer emitted as flat top-level peers, eliminating duplicated content that could mislead LLMs into proposing destructive `suggest_edit` deletions targeting apparent "duplicates". Adds `flattenNodes()` helper for internal error-message callsites.

- **#448 — frontmatter in suggest_edit**: When Claude sends content with `---` frontmatter delimiters via `suggest_edit`, the block previously rendered as a thematic break or heading in TipTap. Now `executeSuggestEdit` detects frontmatter, applies it to the doc via `setFrontmatter()`, and inserts only the body into the editor.

- **#457 — MCP suggestion popup author**: The `onMcpToolInvoke` handler in `App.tsx` was calling `executeTool` with no provenance, so the suggestion popup displayed "unknown" as the author. Now passes `{ model: 'Claude (MCP)' }` plus the live `documentId` and `requestId` so the extension displays the correct label.

- **#455 — new comment MCP tools**: Three new tools exposed via MCP for Claude Desktop:
  - `list_comments` — returns all comments as `{ id, markedText, comment, createdAt, from, to }[]`
  - `add_comment` — accepts `nodeId` or `from`/`to` range + comment text; returns new comment `id`
  - `resolve_comment` — removes a comment by `id`
  All three wire to the existing `getComments` / `setComment` / `unsetComment` from the comments extension without modifying it.

## Test plan

- [ ] Open a doc with a blockquote or list; call `read_document` from Claude Desktop — verify container nodes have `children[]` and children are not duplicated at the top level
- [ ] Via MCP `suggest_edit`, send content starting with a `---` frontmatter block — verify frontmatter appears in the FrontmatterEditor, body rendered normally in TipTap
- [ ] Accept a suggestion made via MCP — verify the suggestion popup label reads "Claude (MCP)" not "unknown"
- [ ] Call `list_comments` from Claude Desktop on a doc with comments — verify all comments returned
- [ ] Call `add_comment` with a `nodeId` — verify comment mark appears on that node
- [ ] Call `resolve_comment` with a comment id — verify the comment mark is removed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #458, opened 2026-05-09
MCP read_document flattens container children as top-level peers


mrangelmarino:
## Summary

The `nodes` array returned by `Prose:read_document` includes children of container nodes (blockquotes, lists) flattened into the top-level array as peers of the container itself. There is no indication of the parent-child relationship. This causes the same logical content to appear in the response twice — once as the container's serialized text content, once as the child's own entry — making consumers (especially LLMs) interpret the structure as containing duplicate content that doesn't actually exist in the document.

## Severity note

Beyond being cosmetically confusing, this bug actively misleads downstream tooling. An LLM consumer working from `read_document` output will reasonably conclude that the apparent duplicates are real artifacts and propose `suggest_edit` deletions targeting the "duplicate" node IDs. Those deletions would in fact empty the inner content of the container, not delete a duplicate. So the bug can produce data-destroying suggestions if accepted by a user trusting the tool's reading.

## What I observed

A document containing this markdown:

```markdown
> "Some quoted text here."
```

returns a `nodes` array containing two consecutive entries:

```json
{
  "id": "h10znllu",
  "type": "blockquote",
  "content": "\"Some quoted text here.\""
},
{
  "id": "y8xvu83b",
  "type": "paragraph",
  "content": "\"Some quoted text here.\""
}
```

Both have identical `content`. There is only one blockquote in the actual rendered markdown.

## Most striking repro

A blockquote with structured children — paragraph + bulletList + paragraph:

```markdown
> Extrapolated to full Dec–May:
>
> - Standard API: ~$6,500
> - Enterprise: ~$26,000
>
> Vs what you actually paid: ~$500.
>
> - Subsidy at standard API: ~13×
> - Subsidy at 4× enterprise: ~50×
```

This single blockquote appears in the `nodes` array as **7 peer entries**:

1. The `blockquote` container (with its full text content concatenated)
2. A `paragraph` ("Extrapolated to full Dec–May:")
3. A `listItem` ("Standard API: \~$6,500")
4. A `listItem` ("Enterprise: \~$26,000")
5. A `paragraph` ("Vs what you actually paid: \~$500.")
6. A `listItem` ("Subsidy at standard API: \~13×")
7. A `listItem` ("Subsidy at 4× enterprise: \~50×")

[...]

PR #495, opened 2026-05-15
fix: post-merge bug batch (#489, #490, #491, #492)

### LINKED ISSUE #489: read_document: off-by-one in collectNodes defeats list-item de-dup (regression of #458)
## Problem

The \`collectNodes\` walker added in #479 / commit \`6f89bf6\` has an off-by-one in its position arithmetic that re-introduces exactly the duplication #458 set out to eliminate — just nested under the listItem instead of as flat top-level peers.

\`src/renderer/extensions/node-ids/index.ts:167\`:

\`\`\`ts
const childPos = pos + offset + 1 // +1 for the parent's opening token
\`\`\`

The \`+1\` is correct for nested containers (a blockquote at P has its content at P+1), but the root \`doc\` has no opening token in ProseMirror position math. The entry call \`collectNodes(doc, 0, doc)\` therefore yields \`childPos = 1\` for the first top-level child instead of \`0\`, and the error propagates downstream.

## User-visible consequence

The listItem-skip filter at \`node-ids/index.ts:184-189\` is the load-bearing user-visible victim:

\`\`\`ts
if (child.type.name === 'paragraph') {
  const parent = doc.resolve(childPos).parent
  if (parent.type.name === 'listItem' || parent.type.name === 'taskItem') {
    return
  }
}
\`\`\`

With \`childPos\` shifted +1, \`doc.resolve(childPos)\` lands *inside* the paragraph rather than between \`listItem\`-open and \`paragraph\`-open, so \`.parent\` resolves to \`paragraph\`. The skip never fires.

Result: every \`listItem\` in \`read_document\` output ends up with a child paragraph whose \`content\` duplicates the listItem's own \`textContent\` — the exact \`{ content: "x", children: [{ content: "x" }] }\` pattern #458 was filed to fix.

## Repro

1. Open a doc with a bullet list (any markdown with \`- item\`).
2. From Claude Desktop call \`read_document\`.
3. Inspect a listItem entry's \`children[]\` — observe an inner paragraph node with content identical to the listItem's own \`content\`.

## Fix options

**A.** Pass \`-1\` as the initial \`pos\`:
\`\`\`ts
return collectNodes(doc, -1, doc)
\`\`\`
\`-1 + offset + 1\` evaluates to \`offset\` for the root, matching \`doc.descendants\` semantics.

**B.** Restructure to track content-start instead of node-pos:
\`\`\`ts
function collectNodes(node, contentStart, doc) {
  node.forEach((child, offset) => {
    const childPos = contentStart + offset
    // ...recurse with childPos + 1 as new contentStart
  })
}
collectNodes(doc, 0, doc)
\`\`\`

**C.** Branch on \`node === doc\` and skip the \`+1\` when at the root.

Option A is the minimal patch.

## Recommended additional safety

A unit test on \`getNodesWithIds\` with a doc-containing-list fixture asserting \`listItem.children.length === 0\` (when the listItem only contains a single paragraph) would catch this and prevent recurrence.

## Refs

- #458 — the original duplication issue this PR was supposed to fix
- #479 — merged PR introducing the regression (commit \`6f89bf6\`)
- Three independent reviewers flagged this; preserving review record for context

### LINKED ISSUE #490: executeSuggestEdit silently drops body when content has --- delimiters but no real frontmatter
## Problem

The frontmatter-strip block in \`executeSuggestEdit\` assigns \`content = body\` **outside** the \`Object.keys(frontmatter).length > 0\` guard. When the LLM sends content that starts with \`---\` but whose YAML doesn't parse to an object (e.g., bare string, invalid YAML, or a deliberately-authored thematic break), the body still gets the leading \`---\n...\n---\n\` block silently stripped — without any frontmatter being applied.

\`src/renderer/lib/tools/executors/editor.ts:320-326\` (current main, after #479 merge):

\`\`\`ts
if (content.trimStart().startsWith('---')) {
  const { content: body, frontmatter } = parseMarkdown(content)
  if (Object.keys(frontmatter).length > 0) {
    useEditorStore.getState().setFrontmatter(frontmatter)
  }
  content = body  // ← runs even when frontmatter is empty
}
\`\`\`

## Repro

LLM calls \`suggest_edit\` with content like:

\`\`\`
---
some prose
---
real body
\`\`\`

\`parseMarkdown\` matches the regex but \`yaml.load("some prose")\` returns the string \`"some prose"\` — not an object. \`frontmatter\` is \`{}\`, \`setFrontmatter\` is skipped (correct), but \`content\` is now just \`"real body"\` — the \`---\nsome prose\n---\` block is silently lost.

Similar case: deliberately-authored thematic break followed by prose, like a Markdown horizontal rule the LLM included as part of the visible content.

## Fix

Move \`content = body\` inside the keys-non-empty branch:

\`\`\`ts
if (content.trimStart().startsWith('---')) {
  const { content: body, frontmatter } = parseMarkdown(content)
  if (Object.keys(frontmatter).length > 0) {
    useEditorStore.getState().setFrontmatter(frontmatter)
    content = body
  }
}
\`\`\`

## Refs

- #448 — original suggest_edit + frontmatter bug
- #479 — merged PR introducing the silent-drop regression
- Flagged by claude-bot reviews on #479

### LINKED ISSUE #491: Comment MCP tools (add_comment, resolve_comment) bypass isEditorReadOnly()
## Problem

\`executeEdit\`, \`executeInsert\`, and \`executeSuggestEdit\` in \`src/renderer/lib/tools/executors/editor.ts\` all call \`isEditorReadOnly()\` and bail with \`EDITOR_READ_ONLY\` when the editor is in reMarkable-OCR mode or a preview tab.

The new \`executeAddComment\` and \`executeResolveComment\` in \`src/renderer/lib/tools/executors/document.ts\` (added in #479 for #455) also mutate the document (they apply/remove a comment mark via \`setComment\`/\`unsetComment\`), but never call \`isEditorReadOnly()\` — the helper isn't even exported from \`editor.ts\`.

## User-visible consequence

An MCP client (Claude Desktop) can add or remove comments on:
- A reMarkable doc opened in read-only OCR view
- A preview tab (single-click in file explorer)

…while every other mutation tool correctly refuses. The behavior is inconsistent and lets MCP callers bypass the read-only contract.

## Fix

Export \`isEditorReadOnly()\` from \`editor.ts\` (or duplicate the check — it's a 2-line function) and gate the two comment tools the same way as the edit tools:

\`\`\`ts
export function executeAddComment(...) {
  const editor = getEditor()
  if (!editor) return toolError('Editor not available', 'EDITOR_NOT_AVAILABLE')
  if (isEditorReadOnly()) return toolError('Document is read-only in this mode', 'EDITOR_READ_ONLY')
  // ...
}
\`\`\`

Same for \`executeResolveComment\`.

## Refs

- #455 — original MCP comment tools spec
- #479 — merged PR (commit \`6f89bf6\`) where the tools landed without the guard
- Flagged by claude-bot reviews on #479

### LINKED ISSUE #492: CommentPopover doesn't guard against null data-comment-id
## Problem

\`CommentPopover\` (\`src/renderer/components/CommentPopover.tsx\`) reads the comment id from the clicked DOM node without checking for null:

\`\`\`tsx
const id = mark.getAttribute('data-comment-id')
const text = mark.getAttribute('data-comment') || ''

if (id) {
  // open popover
}
\`\`\`

The \`if (id)\` actually does guard the popover-open path, so this is **almost** safe. But the Comment extension's \`renderHTML\` for the \`id\` attribute returns \`{}\` when \`attributes.id\` is falsy — meaning the rendered span may carry class \`comment-mark\` but no \`data-comment-id\`. Clicking such a malformed mark short-circuits silently with no user feedback.

## User-visible consequence

Low-likelihood (well-formed marks always have IDs), but if a mark is ever rendered without one (legacy data, partial extension upgrade, manually-constructed HTML), the click does nothing visible — including no console warning to aid debugging.

## Fix

Add an explicit early return + dev warning:

\`\`\`tsx
const id = mark.getAttribute('data-comment-id')
if (!id) {
  console.warn('[CommentPopover] comment-mark clicked with no data-comment-id', mark)
  return
}
const text = mark.getAttribute('data-comment') || ''
// open popover unconditionally now that id is non-null
\`\`\`

## Refs

- #456 — original comment-click regression
- #480 — merged PR (commit \`2e44a8b\`)
- Flagged by claude-bot reviews on #480

mrangelmarino:
## Summary

Four bugs surfaced via post-merge claude-bot reviews on PRs #479 and #480 (yesterday's Do First sweep). All four are small, scoped, and bundled here so 1.1.1 can ship clean.

| Bug | Severity | Fix |
|---|---|---|
| #489 | **Critical** — defeats #458 | `collectNodes(doc, -1, doc)` so root-level positions don't shift +1. Without this, the paragraph-in-listItem skip never fires and every listItem ends up with a duplicate inner paragraph in `read_document` output. |
| #490 | Data loss | Move `content = body` inside the keys-non-empty guard in `executeSuggestEdit` so malformed YAML doesn't silently strip a real body. |
| #491 | Privilege bypass | Export `isEditorReadOnly` from `editor.ts` and gate `executeAddComment` / `executeResolveComment` with it — they were the only mutation tools bypassing the read-only check. |
| #492 | Defensive | Surface `console.warn` instead of silent no-op when a comment-mark is clicked without a `data-comment-id`. |

## Files

- `src/renderer/extensions/node-ids/index.ts` — #489
- `src/renderer/lib/tools/executors/editor.ts` — #490 + export `isEditorReadOnly`
- `src/renderer/lib/tools/executors/document.ts` — #491
- `src/renderer/components/CommentPopover.tsx` — #492

4 commits, one per bug. Net ~40 lines.

### #489 — read_document nesting

1. Open a doc with a bullet list (e.g., `- one\n- two\n- three`).
2. From Claude Desktop: *"Call the prose read_document tool."*
3. **Expected**: each listItem has `children: []` (or `children` absent for empty), NOT `children: [{ type: "paragraph", content: "<same text as listItem>" }]`. Verify by inspecting the JSON response in Claude Desktop's tool-call output.

### #490 — frontmatter body drop

1. From Claude Desktop: *"Use suggest_edit with content `---\nplain text\n---\nreal body` targeting the first paragraph."*
2. **Expected**: the suggestion popup shows the FULL replacement text including the `---\nplain text\n---` block. No frontmatter is applied to the doc.
3. Regression check — happy path: *"Use suggest_edit with content `---\ntitle: Test\n---\nbody`"* — should still apply frontmatter and show only `body` in the diff.

### #491 — comment tools on read-only

1. Open a reMarkable-synced notebook (read-only OCR view) OR a preview tab (single-click any file from the explorer — don't double-click).
2. From Claude Desktop: *"Use add_comment with this nodeId [any id from read_document] and comment 'test'."*
3. **Expected**: error `EDITOR_READ_ONLY` — *"Document is read-only in this mode"*.
4. Same for `resolve_comment`.

### #492 — null comment-id guard

1. Regression check — add a comment normally, click it. Popover opens.
2. In DevTools console:
   ```js
   const el = document.createElement('span')
   el.className = 'comment-mark'
   document.body.appendChild(el)
   el.click()
   ```
3. **Expected**: console warning `[CommentPopover] comment-mark clicked with no data-comment-id`, no popover opens. Clean up: `el.remove()`.

## Workflow constraint

**Do NOT auto-merge.** This batch ships before 1.1.1 — needs eyes-on confirmation that #489 actually undoes the listItem duplication and that #491's gate works on a real reMarkable doc. Hand off to user for manual verification.

Closes #489
Closes #490
Closes #491
Closes #492

### DIFF
--- a/src/renderer/components/CommentPopover.tsx
+++ b/src/renderer/components/CommentPopover.tsx
@@ -40,17 +40,21 @@ export function CommentPopover({ editor }: CommentPopoverProps) {
         event.stopPropagation()
 
         const id = mark.getAttribute('data-comment-id')
-        const text = mark.getAttribute('data-comment') || ''
-
-        if (id) {
-          const rect = mark.getBoundingClientRect()
-          setPopover({
-            isOpen: true,
-            commentId: id,
-            commentText: text,
-            position: { x: rect.left + rect.width / 2, y: rect.bottom + 8 },
-          })
+        if (!id) {
+          // Malformed mark — the Comment extension's renderHTML drops the
+          // attribute when attrs.id is falsy. Warn loudly so the source of
+          // the bad mark is debuggable, but don't silently swallow the click.
+          console.warn('[CommentPopover] comment-mark clicked with no data-comment-id', mark)
+          return
         }
+        const text = mark.getAttribute('data-comment') || ''
+        const rect = mark.getBoundingClientRect()
+        setPopover({
+          isOpen: true,
+          commentId: id,
+          commentText: text,
+          position: { x: rect.left + rect.width / 2, y: rect.bottom + 8 },
+        })
         return
       }
 
--- a/src/renderer/extensions/node-ids/index.ts
+++ b/src/renderer/extensions/node-ids/index.ts
@@ -213,7 +213,13 @@ function collectNodes(
 export function getNodesWithIds(
   doc: import('@tiptap/pm/model').Node
 ): NodeWithId[] {
-  return collectNodes(doc, 0, doc)
+  // pos = -1 so the first top-level child's childPos = -1 + 0 + 1 = 0, matching
+  // ProseMirror's content-position semantics for the doc root (which, unlike
+  // inner container nodes, has no opening token to skip). Without this,
+  // doc.resolve(childPos) lands one position too deep inside top-level paragraphs
+  // and the listItem-skip filter below never fires — re-introducing the
+  // duplication this walker was rewritten to eliminate. See #458, #489.
+  return collectNodes(doc, -1, doc)
 }
 
 /**
--- a/src/renderer/lib/tools/executors/document.ts
+++ b/src/renderer/lib/tools/executors/document.ts
@@ -12,6 +12,7 @@ import { getNodesWithIds, findNodeById } from '../../../extensions/node-ids'
 import type { NodeWithId } from '../../../extensions/node-ids'
 import { getComments } from '../../../extensions/comments'
 import { getAISuggestions } from '../../../extensions/ai-suggestions'
+import { isEditorReadOnly } from './editor'
 import { getApi } from '../../browserApi'
 import { generateId } from '../../persistence'
 
@@ -379,6 +380,10 @@ export function executeAddComment(args: {
     return toolError('Editor not available', 'EDITOR_NOT_AVAILABLE')
   }
 
+  if (isEditorReadOnly()) {
+    return toolError('Document is read-only in this mode', 'EDITOR_READ_ONLY')
+  }
+
   const { nodeId, comment } = args
   let from = args.from
   let to = args.to
@@ -440,6 +445,10 @@ export function executeResolveComment(args: {
 
   if (!editor) {
     return toolError('Editor not available', 'EDITOR_NOT_AVAILABLE')
+  }
+
+  if (isEditorReadOnly()) {
+    return toolError('Document is read-only in this mode', 'EDITOR_READ_ONLY')
   }
 
   const { id } = args
--- a/src/renderer/lib/tools/executors/editor.ts
+++ b/src/renderer/lib/tools/executors/editor.ts
@@ -27,9 +27,10 @@ function getEditor(): Editor | null {
 
 /**
  * Check if the editor is in a read-only mode (reMarkable OCR or preview tab).
- * AI tools should not mutate the document in these modes.
+ * AI tools should not mutate the document in these modes. Exported so that
+ * document-tool executors (add_comment, resolve_comment) can apply the same gate.
  */
-function isEditorReadOnly(): boolean {
+export function isEditorReadOnly(): boolean {
   const state = useEditorStore.getState()
   return state.isRemarkableReadOnly || state.isPreviewTab
 }
--- a/src/renderer/lib/tools/executors/editor.ts
+++ b/src/renderer/lib/tools/executors/editor.ts
@@ -320,9 +320,15 @@ export function executeSuggestEdit(
   if (content.trimStart().startsWith('---')) {
     const { content: body, frontmatter } = parseMarkdown(content)
     if (Object.keys(frontmatter).length > 0) {
+      // Real frontmatter — apply it to the store and use the stripped body
+      // as the suggestion content.
       useEditorStore.getState().setFrontmatter(frontmatter)
+      content = body
     }
-    content = body
+    // If parseMarkdown matched the regex but yielded no keys (e.g., bare
+    // string YAML, malformed YAML, or a deliberate thematic break followed
+    // by prose), leave content untouched. Otherwise we'd silently drop the
+    // ---...--- block from the suggestion. See #490.
   }
 
   // Find the node by ID, fall back to content matching if stale


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/0ef7bf8d-6d80-481b-ad20-c2b2e4894b0c?refs=aa0ce365d6a8d4d0cb39e7b32117c73c768a4997&source=prChecksComment) |

mrangelmarino:
/review 

PR #507, opened 2026-05-16
fix(mcp): node-id walker bugs in list items and container wrappers (#496, #497)

### LINKED ISSUE #496: List items share the same nodeId across siblings (read_document returns dupes)
## Problem

Pasting a simple bullet list and calling \`read_document\` returns three listItems with the *same* \`nodeId\`:

\`\`\`json
{
  "type": "bulletList",
  "children": [
    { "id": "zn2okfv2", "type": "listItem", "content": "One" },
    { "id": "zn2okfv2", "type": "listItem", "content": "Two" },
    { "id": "zn2okfv2", "type": "listItem", "content": "Three" }
  ]
}
\`\`\`

\`content\` differs (so the listItems *are* distinct nodes), but every listItem reads back \`zn2okfv2\` as its id.

## Severity

\`read_document\`'s schema says *"each node with a unique ID"* and downstream tools (\`suggest_edit\`, \`add_comment\`, \`resolve_comment\`) use \`nodeId\` as the primary targeting key. With duplicates:

- An LLM targeting any of the three listItems by id always lands on the **first one** (via \`findNodeById\` short-circuiting on first match).
- The \`search\` fallback would have to do all the work — which is exactly the brittle path \`nodeId\` was added to avoid.

## Suspected root cause

The pattern (one shared id across all siblings of the same type, in the same parent) is consistent with the \`attrs\` object being reference-shared across listItems coming out of the markdown parser, so when the \`nodeIds\` extension's \`appendTransaction\` assigns one id, the walker sees the same shared attrs reference for the other two.

Worth investigating:
- \`tiptap-markdown\`'s listItem parser — does it produce nodes whose attrs are copies-of-default or shared singletons?
- The \`NodeIds\` extension's \`appendTransaction\` (\`src/renderer/extensions/node-ids/index.ts:47-83\`) — the check \`if (node.attrs.nodeId) return\` would short-circuit subsequent listItems if their attrs reference the same (already-updated) object.
- ProseMirror's \`setNodeMarkup\` behavior when given attrs spread from a shared reference — probably correct, but verify under the actual code path.

## Repro

1. New Untitled tab.
2. Paste \`- one\n- two\n- three\`.
3. From Claude Desktop: \`Call the prose read_document tool.\`
4. Observe identical \`id\` on all listItem children.

## Related

- #489 — fixed off-by-one duplication (children-as-paragraphs) but did NOT touch id assignment; this is a separate upstream issue
- #458 — original \`read_document\` correctness work

### LINKED ISSUE #497: Container wrappers (bulletList/orderedList/taskList) emit nodeId: '' in read_document
## Problem

\`read_document\` emits container wrapper nodes with an empty-string id:

\`\`\`json
{
  "id": "",
  "type": "bulletList",
  "content": "OneTwoThree",
  "children": [...]
}
\`\`\`

## Why

The walker's \`CONTAINER_TYPES\` set in \`src/renderer/extensions/node-ids/index.ts:131-139\` includes \`bulletList\`, \`orderedList\`, \`taskList\`, but the \`NODE_TYPES_WITH_IDS\` set the \`NodeIds\` extension uses to assign IDs (lines 17-24) does **not**. So:

1. The extension never assigns IDs to list wrappers (by design — they're structural).
2. But the walker emits them anyway (because they're in \`CONTAINER_TYPES\`) with a \`nodeId: child.attrs.nodeId || ''\` fallback.

Result: container wrappers ship with \`id: ""\` to MCP consumers.

## Severity

Footgun, not catastrophic. The empty string:
- Collides on first-match if anything ever does \`nodes.find(n => n.id === '')\`.
- Looks like a valid id to an LLM consumer (which might try to target it via \`suggest_edit\`).
- The new \`read_document\` description says container nodes carry a \`children\` array, but doesn't tell the LLM \"don't target the wrapper.\"

The PR #479 review thread flagged this as a minor concern. Worth fixing for 1.1.x.

## Fix options

**A.** Add \`bulletList\`, \`orderedList\`, \`taskList\` to \`NODE_TYPES_WITH_IDS\` so they get real IDs. Smallest change. Consumers can target the wrapper as an addressable container.

**B.** Drop empty-id container entries from the walker output. The container's children would still surface (their de-dup is the point of the nesting refactor), but the wrapper itself wouldn't be a top-level entry. May complicate consumers that walk \`children\`.

**C.** Keep emitting but use a sentinel \`null\` or omit the \`id\` field entirely for un-id'd containers. Type the schema accordingly so consumers know not to target.

Option A is the cleanest IMO and matches how list extensions usually work in other editors.

## Repro

Same as #496 — paste a bullet list, call \`read_document\`, observe the wrapper's \`id: ""\`.

## Related

- #489 — off-by-one fix in the same walker
- #458 — original \`read_document\` correctness work
- post-merge review of #479 (PR conversation comments) flagged this

mrangelmarino:
## Summary

- **#496**: Sibling `listItem` nodes shared the same `nodeId` because `tiptap-markdown` can produce list nodes with a reference-shared `attrs` object. After `setNodeMarkup` wrote an id to the first sibling, the shared attrs object reflected that id — so the `if (node.attrs.nodeId)` guard short-circuited all subsequent siblings, leaving them all with the same id. Fixed by reading from `tr.doc.nodeAt(pos)` (the transaction's own document state) instead of `node.attrs`, so each sibling is checked independently.
- **#497**: `bulletList`, `orderedList`, `taskList` were in `CONTAINER_TYPES` (emitted by the walker) but absent from `NODE_TYPES_WITH_IDS` (so the extension never assigned them ids). Every list wrapper arrived at MCP consumers with `id: ""`. Fixed by adding the three types to `NODE_TYPES_WITH_IDS` (Option A from the issue).

## Test plan

- [ ] `npm run build` passes (verified: clean build)
- [ ] Open a doc with a bullet list; call `read_document` — each `listItem` should have a unique non-empty `id`
- [ ] Verify `bulletList` / `orderedList` / `taskList` wrappers also have non-empty `id` values
- [ ] Verify `suggest_edit` targeting any list item lands on the correct item (not always the first)

Closes #496
Closes #497

### DIFF
--- a/src/renderer/extensions/node-ids/index.ts
+++ b/src/renderer/extensions/node-ids/index.ts
@@ -19,6 +19,9 @@ const NODE_TYPES_WITH_IDS = [
   'heading',
   'codeBlock',
   'blockquote',
+  'bulletList',
+  'orderedList',
+  'taskList',
   'listItem',
   'taskItem',
 ]
@@ -65,8 +68,16 @@ export const NodeIds = Extension.create({
             // Only process nodes that should have IDs
             if (!NODE_TYPES_WITH_IDS.includes(node.type.name)) return
 
-            // Skip if node already has an ID
-            if (node.attrs.nodeId) return
+            // Check the transaction's current mapped document for an existing ID.
+            // We intentionally do NOT use `node.attrs.nodeId` here: when tiptap-markdown
+            // parses a list, sibling listItems may share the same attrs object reference.
+            // After `setNodeMarkup` assigns an id to the first sibling, that shared attrs
+            // object reflects the new id — so `node.attrs.nodeId` on subsequent siblings
+            // would appear truthy even though no id was written for them, causing all
+            // siblings to report the same id. Reading from `tr.doc` gives us the actual
+            // committed state for each node individually.
+            const currentNode = tr.doc.nodeAt(pos)
+            if (currentNode?.attrs.nodeId) return
 
             // Assign a new ID
             tr.setNodeMarkup(pos, undefined, {


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/86dca181-52b9-4d07-bd84-b59b23fea23f?refs=d51866c92b187a53ca58e3411254ab7c4faf3ad0&source=prChecksComment) |

mrangelmarino:
/review 

PR #513, opened 2026-05-16
feat(#467): persona seed + StatusBar mode labels (Chat / Editor / Create)

mrangelmarino:
## Summary

Chunk 1 of the #467 umbrella (built-in chat agent persona + chat-mode refactor). Rewrites the system prompt to enforce an editor-not-author posture and updates the StatusBar mode labels to match what each mode actually does.

Refs #467, #468.

## Intentional transitional state

Constants keep their legacy names (`SUGGESTIONS_MODE_INSTRUCTIONS` / `PLAN_MODE_INSTRUCTIONS` / `FULL_MODE_INSTRUCTIONS`) until Chunk 3 renames them after #450 merges. Behavior is correct; identifiers lag. Same for the `ToolMode` union (`suggestions / plan / full`) and `modeConfig` keys in `StatusBar.tsx`.

### `src/renderer/lib/prompts.ts`

`BASE_PROMPT` is rewritten to lead with editor-not-author posture and embed the persona heuristics from the #468 seed (push once, surface the user's crisp lines, reflect words accurately, structural diagnosis before scaffolding) plus the escape-hatch protocol. The "When asked to write or draft new content, output markdown directly in your response" rule is gone — drafting now requires Create Mode.

The three mode-instructions constants are reshaped to match Chat / Editor / Create posture:

- **`SUGGESTIONS_MODE_INSTRUCTIONS`** (Chat) — sounding board, fact-check, pushback. Still no tools wired in this mode; Chunk 2 of #467 adds read-only tool wiring.
- **`PLAN_MODE_INSTRUCTIONS`** (Editor, default) — propose copy edits via `suggest_edit`, leave editorial notes via `add_comment`, never author into the document. Includes the `suggest_edit` vs `add_comment` discrimination heuristic from #468.
- **`FULL_MODE_INSTRUCTIONS`** (Create, opt-in) — no-authorship rule lifted, but persona constraints (accuracy, structural diagnosis, concision) still apply. Adds the "prefer `suggest_edit` over `edit`/`insert` for existing authored content; `edit`/`insert` for drafting or unambiguous fixes" guidance.

### `src/renderer/components/layout/StatusBar.tsx`

`modeConfig` labels updated from the misleading "suggest edits" / "plan" / "accept edits" to **chat** / **editor** / **create** with descriptions that match the new posture. Internal keys unchanged.

## Persona heuristics, quoted inline from #468

So reviewers see the prompt delta in this PR, not by clicking through:

### Heuristics (all modes)

[...]

Issue #468, opened 2026-05-12
Built-in agent persona + chat-mode refactor (Editor / Create / Chat)


mrangelmarino:
Part of #467.

## Scope

Three interlocking pieces of work:

1. **Rewrite the built-in chat agent's system prompt** to enforce an editor-not-author posture by default.
2. **Refactor the chat-mode names** from capability-flavored (`suggestions` / `plan` / `full`) to intention-flavored (**Chat** / **Editor** / **Create**), and reshape each mode's behavior to match the new posture (see "Mode behaviors" below).
3. **Depend on the `add_comment` tool** delivered by #455 (comment capabilities — MCP and built-in agent). Editor Mode's posture (propose + annotate, never author) requires this tool to exist so the agent has a non-drafting way to flag issues without proposing a specific replacement.

Per #467: Create Mode is **toggleable** by the user. Editor Mode is the default.

### Mode behaviors (target)

| Mode | Tools available | Authors directly? |
|---|---|---|
| **Chat** | Read-only: `read_document`, `get_outline`, `read_selection`, `search_document`, `get_metadata` | No |
| **Editor** *(default)* | All Chat tools + `suggest_edit` (copy edits) + `add_comment` (editorial notes) | No — proposes concrete changes and annotates with judgment-bearing notes |
| **Create** | All Editor tools + `edit` + `insert` + drafting allowed | Yes — opt-in |

**`suggest_edit` vs `add_comment`:**
- `suggest_edit` = user-sanctioned copy edit. The agent proposes the exact replacement text; the user reviews and accepts/rejects. Use for typos, formatting, mechanical fixes — anywhere there's a clear right answer.
- `add_comment` = editorial note. The agent flags an issue without proposing the replacement prose. Use for judgment-bearing direction like "tighten this paragraph," "this conflicts with paragraph 2," "needs a transition." Proposing specific prose changes here would cross into authorship.

### Files to modify

- `src/renderer/lib/prompts.ts` — rewrite `BASE_PROMPT` (lines 21-44), the three `*_MODE_INSTRUCTIONS` blocks (46-77), and `buildSystemPrompt()` (79-115)
- `src/shared/tools/types.ts:11` — rename `ToolMode` union values
- `src/renderer/stores/chatStore.ts:18` — duplicate `ToolMode` declaration (consolidate against `shared/tools/types.ts`)
- `src/shared/tools/registry.ts:43-54` — mode-gating logic in `checkToolAccess` / `getDefaultMode`
[...]

Issue #467, opened 2026-05-12
Built-in chat agent: persona, modes, and tool-surface coherence


mrangelmarino:
## Strategic positioning

**Hypothesis:** Prose is the AI editor that doesn't fuck with your voice.

Other AI editors draft prose by default. Prose differentiates by treating the writer's voice as inviolable and the AI as an editor in the original sense — scaffolding, questioning, fact-checking, sounding-board feedback. Authoring stays with the human.

This umbrella covers the work to make the built-in chat agent (the one users talk to inside the app — *not* the MCP server) live up to that hypothesis, plus coupled work on chat-mode semantics, model handling, and MCP correctness bugs that affect the same surface.

## Configurability — settled

The "no drafting" default ships as **Option B (toggleable)**: Editor Mode is the default; users can toggle into Create Mode to opt in to LLM-authored prose. Same brand asymmetry — Cursor/Notion can't claim "your voice is sacred by default" even with a toggle — but flexible for users who want a draft starting point.

Open follow-up: the exact UI affordance (how prominent / discoverable Create Mode is) is a design call that lives with the sub-issue.

## Proposed mode refactor

Current chat-agent modes (`src/shared/tools/types.ts:11`) describe capabilities, not intentions, and the user-facing labels in `src/renderer/components/layout/StatusBar.tsx:47-53` are actively misleading:

| Internal | UI label | What it actually does |
|---|---|---|
| `suggestions` | "suggest edits" | No tools at all — pure chat feedback |
| `plan` | "plan" | Read + `suggest_edit` (user-reviewed proposals) |
| `full` | "accept edits" | Adds `edit` / `insert` (direct overwrites) |

Proposed rename to intention-named modes:

| Current | Proposed | Posture |
|---|---|---|
| `suggestions` | **Chat Mode** | Read-only tools, no edits — conversation, fact-check, sounding board. The agent can read the document, but cannot modify it. |
| `plan` | **Editor Mode** *(default)* | All Chat Mode read-only tools + `suggest_edit` + `add_comment`. The agent proposes concrete copy edits via `suggest_edit` (typos, formatting, mechanical fixes the user can accept) and leaves editorial notes via `add_comment` (judgment-bearing direction like "tighten this paragraph"). Never authors directly. |
| `full` | **Create Mode** | All Editor Mode tools + `edit` + `insert` + drafting allowed. The LLM is permitted to author prose into the document. Persona constraints around accuracy and concision still apply. **User-toggleable.** |

[...]

Issue #455, opened 2026-05-09
MCP: surface comment capabilities for read/add/resolve


mrangelmarino:
## Problem

Prose has a comments feature in the editor (`src/renderer/extensions/comments/`) — TipTap marks with `id`, `comment`, `createdAt` that highlight a range of text and attach a note. It's exposed in-app for AI-assisted editing flows but **not surfaced through MCP**, so Claude can't read what comments exist or add new ones from a conversation.

## What we have today

- `Comment` TipTap mark + `getComments(editor)` helper return all comments as `CommentData[]` (id, markedText, comment, createdAt, from, to).
- Renderer-side commands for adding/removing comments via the editor instance.
- No corresponding MCP tools.

## Proposed MCP tools

Rough sketch — granularity TBD during implementation:

- `list_comments` — return all comments in the active document, including the marked text excerpt, the comment body, and the node id (or position range) they're attached to. Useful for Claude to summarize feedback, address review comments, or find related ones before adding new.
- `add_comment` — attach a comment to a target. Params: `nodeId` (or `from`/`to` range, mirroring the `suggest_edit` shape), `comment` text. Returns the new comment id.
- `resolve_comment` — remove a comment by id. ('Resolve' framing matches the conversational verb users actually use.)

Open questions:

- Should `add_comment` require a `nodeId` like `suggest_edit`, or accept a text-search fallback?
- Do we expose a `reply_to_comment` (creating a thread)? Today comments are flat; threading would be a separate feature.
- Author attribution: comments don't currently track author. If Claude adds one, do we tag it? (Probably yes — `author: 'claude'` field on the mark, displayed in the UI.)

## Why this matters

Two concrete flows it unlocks:

1. **Comment-driven editing.** *"Address the comments in my doc"* → Claude calls `list_comments`, walks each one, calls `suggest_edit` to address it, then `resolve_comment` after the user accepts the diff.
2. **Conversational annotation.** Claude offers a critique while reading via `read_document` — *"want me to leave that as a comment?"* → `add_comment` lands a margin note the user can revisit later in Prose.

## Out of scope

- Comment threading / replies.
- Cross-document comment search (this is per-document).
- Comment notifications / sync.

## Refs

- #435 — Prose-in-Claude umbrella; this is a natural follow-on capability once the editor widget + MCP create/edit flows are stable.
- Existing MCP tools: `read_document`, `suggest_edit`, `get_outline`, `open_file`, `create_and_open_file`.
[...]

PR #514, opened 2026-05-16
feat(#467): Chat Mode passes read-only tools instead of nothing

mrangelmarino:
## Summary

Chunk 2 of the #467 umbrella. Chat Mode (internal `ToolMode` key `'suggestions'`) currently passes zero tools to the LLM — `useChat.ts:306,484` had a `state.toolMode !== 'suggestions' ? getToolsForClaudeAPI(...) : undefined` ternary. This means the in-app chat agent couldn't even read the document to ground its responses. Chunk 2 wires a read-only tool subset so Chat Mode becomes a real sounding-board mode that can ground itself in the doc.

**Stacked on #513** — base branch is `issue-467-persona-seed-prompts`. Merge #513 first.

Refs #467, #468.

### `src/renderer/hooks/useChat.ts`

- New module-level allowlist: `CHAT_MODE_TOOL_NAMES` — `read_document`, `read_selection`, `get_outline`, `search_document`, `get_metadata`, `list_comments`.
- New helper: `getToolsForToolMode(toolMode)` — returns the Chat Mode subset for `'suggestions'`, otherwise delegates to `getToolsForClaudeAPI(toolMode)`.
- Both call sites (`:306`, `:484`) collapse from the old ternary into a single `getToolsForToolMode(...)` call.
- Tool-loop init guard (line ~493) tightened from `if (tools)` to `if (tools && tools.length > 0)` to preserve the original "don't init when there are no tools" semantics now that `tools` is always an array.

### `src/renderer/lib/prompts.ts`

- `SUGGESTIONS_MODE_INSTRUCTIONS` rewritten to describe the new read-only tool surface, with a redirect note: "If the user asks for an edit or annotation, tell them to switch to Editor Mode (StatusBar → editor) and re-ask." The previous version said the agent has no tools and should put proposed revisions in chat text.
- Touched the `// Line references only when no tools available` comment in `buildSystemPrompt()` — now reads "Chat Mode: agent references doc content in plain chat replies; line refs render as clickable jump-to-line links in the UI" since Chat Mode has tools now. The line-refs block itself is kept as-is (clickable jump-to-line is useful regardless of tool surface).

## Deviation from the approved chunking plan — read this

The plan's Chunk 2 bullet said: *"filter by `category === 'document'` when in `suggestions` mode."* That filter is **too loose** in this codebase as it stands — `category: 'document'` includes the **mutating** `add_comment` and `resolve_comment` tools (see `src/shared/tools/schemas/document.ts:142,160`). A pure category filter would silently put `add_comment` in Chat Mode, breaking the "Chat Mode is read-only" invariant the chunking plan explicitly relies on.

This PR substitutes an explicit name allowlist instead. Rationale:

- **It's read-only by construction.** The set lists the six tools that actually read; you can't accidentally let a mutating tool through by typo or by a downstream category change.
- **It survives #450.** When `list_tabs` / `select_tab` land, they'll have a different category anyway (likely `'ui'` or a new tab category). The plan's `category === 'document'` filter would have excluded them automatically; the allowlist explicitly excludes them too, and we revisit when #450 merges to decide whether tab discovery belongs in Chat Mode.
- **The plan's reasoning still holds.** Category-vs-`requiresMode` was the user's preference because `category` is intent-based. An explicit allowlist is also intent-based — just at a finer grain.

If you'd prefer the pure category filter and accept the temporary `add_comment` leak until Chunk 4 sets `requiresMode: 'editor'` on it, say so and I'll swap it. Default stance: stay surgical.

## #450 / #447 contention status

- `gh pr list --repo solo-ist/prose --state open --search '450 list_tabs select_tab'` → no open PR. #450 not yet in flight.
- `src/shared/tools/registry.ts` not touched in this PR (the filter lives at the call site in `useChat.ts`), so this PR does not race with #450.
- `src/renderer/extensions/ai-suggestions/**` not touched, so no race with #447.

## Tests

No existing tests reference `useChat`'s tool wiring, `getToolsForClaudeAPI`, `CHAT_MODE_TOOL_NAMES`, or `SUGGESTIONS_MODE_INSTRUCTIONS` output. Verified via the same grep as Chunk 1 (#513). No test updates needed.

If we want unit coverage for the Chat Mode filter (recommended once we add a vitest setup), the test would assert `getToolsForToolMode('suggestions').map(t => t.name)` matches the allowlist exactly and excludes `add_comment` / `resolve_comment`. Leaving as a follow-up unless you want it inline.

## Verification

1. `npm run dev` — main process + preload + vite all build clean.
2. Open the app on the Chunk 2 branch. Switch to **chat** mode in the StatusBar.
3. Prompt the chat: *"what's in my document?"* — agent calls `read_document`, summarizes the content. Previously: agent had no tools and either guessed or refused.
4. Prompt: *"add a comment on paragraph 2"* — agent declines and tells the user to switch to Editor Mode. (Mutating comment tools are not in the Chat Mode allowlist.)
5. Switch to **editor** mode. Same comment-add request now works.
6. Switch to **create** mode. All tools still available, no regression.

## Out of scope (in this chunk)

- `ToolMode` rename to `chat / editor / create` → Chunk 3 (blocked on #450).
- `add_comment` gated by `requiresMode: 'editor'` → Chunk 4 (after Chunk 3 + #455 verification).
- Accept-flow refactor → Chunk 5 (blocked on #447).

### DIFF
--- a/src/renderer/hooks/useChat.ts
+++ b/src/renderer/hooks/useChat.ts
@@ -1,5 +1,5 @@
 import { useCallback, useEffect } from 'react'
-import { useChatStore, createMessageId } from '../stores/chatStore'
+import { useChatStore, createMessageId, type ToolMode } from '../stores/chatStore'
 import { useSettingsStore } from '../stores/settingsStore'
 import { useEditorStore } from '../stores/editorStore'
 import { useEditorInstanceStore } from '../stores/editorInstanceStore'
@@ -12,6 +12,32 @@ import { executeTool, resolveToolPosition } from '../lib/tools'
 import { getToolsForClaudeAPI } from '../../shared/tools/registry'
 import { resolveModelName } from '../../shared/llm/models'
 import type { LLMMessage, LLMStreamToolCall, LLMContentBlock } from '../types'
+
+// Chat Mode (legacy internal key 'suggestions') gets a read-only tool subset
+// rather than zero tools, so the agent can ground itself in the document
+// without proposing edits. We can't filter by `category === 'document'` alone
+// — that category includes the mutating add_comment and resolve_comment tools.
+// Mutating comment tools move behind Editor Mode via `requiresMode` in Chunk 4
+// of #467; until then this explicit allowlist is the gate.
+//
+// Audit checkpoint: when #450 (list_tabs / select_tab) merges, decide whether
+// those new tools belong in Chat Mode and extend this set, or defer them to
+// Chunk 3.
+const CHAT_MODE_TOOL_NAMES: ReadonlySet<string> = new Set([
+  'read_document',
+  'read_selection',
+  'get_metadata',
+  'search_document',
+  'get_outline',
+  'list_comments'
+])
+
+function getToolsForToolMode(toolMode: ToolMode): ReturnType<typeof getToolsForClaudeAPI> {
+  if (toolMode === 'suggestions') {
+    return getToolsForClaudeAPI('full').filter((t) => CHAT_MODE_TOOL_NAMES.has(t.name))
+  }
+  return getToolsForClaudeAPI(toolMode)
+}
 
 // Module-level flag to ensure stream listeners are only registered once globally
 let streamListenersInitialized = false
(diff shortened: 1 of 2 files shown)


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/86306221-0b9d-4529-9612-8e813a900b9a?refs=fedf80471050a245ca704ecbf46a6deb17dc3460&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
/review 

PR #522, opened 2026-05-17
refactor(#467): rename ToolMode → chat/editor/create + flip default to Editor

mrangelmarino:
## Summary

Chunk 3 of the #467 umbrella. Internal `ToolMode` keys finally catch up to the user-facing labels established in Chunks 1+2 (#513, #521). Editor Mode becomes the **actual** default — fresh users no longer land in Create Mode (legacy `full`).

Refs #467, #468.

## Surface-by-surface

| File | Change |
|---|---|
| `src/shared/tools/types.ts` | `ToolMode` union: `'suggestions' \| 'full' \| 'plan'` → `'chat' \| 'editor' \| 'create'`. |
| `src/shared/tools/registry.ts` | Rewrite `getToolsForMode` / `isToolAvailableInMode` as a clean capability ladder via `MODE_LEVEL` (`chat ⊂ editor ⊂ create`). Drops the special-case branches. `getToolsForClaudeAPI` default param `'full'` → `'create'`. |
| `src/shared/tools/schemas/editor.ts` | `edit` and `insert` `requiresMode: 'full'` → `'create'`. |
| `src/renderer/stores/chatStore.ts` | Drop duplicate `ToolMode` declaration; re-export from `shared/tools/types`. **Flip initial `toolMode` `'full'` → `'editor'`** (the safety-by-default promise of #467). `agentMode` legacy boolean now defaults `false` and is kept in sync with `toolMode === 'create'`. New `cycleToolMode` action cycles chat → editor → create → chat for the Shift+Tab keyboard shortcut. |
| `src/renderer/hooks/useChat.ts` | `CHAT_MODE_TOOL_NAMES` filter check + `maxTokens` ternary updated. Exposes `cycleToolMode`. |
| `src/renderer/lib/tools/modes.ts` | `getDefaultMode()` returns `'editor'`. `checkToolAccess` labels use the new `MODE_LABEL` map. |
| `src/renderer/lib/tools/index.ts` | `executeTool` default `mode: 'full'` → `'create'`. |
| `src/renderer/components/layout/StatusBar.tsx` | `modeConfig` keys updated to new union. **`(default)` marker restored on Editor** (the marker was dropped in #513's fix because the default hadn't actually flipped; now it has). |
| `src/renderer/components/layout/App.tsx` | MCP-mode `executeTool` call: `'full'` → `'create'`. |
[...]

Issue #447, opened 2026-05-01
Caption AI edits with the LLM's reason


mrangelmarino:
**As a user**, when I hover the AI annotation that persists after accepting an AI edit, I want to see the model's reasoning for that edit — preserving the explanation that was shown pre-accept in the suggestion popover.

### Current state

- **Pre-accept** — the suggestion popover renders `Explanation:` with the model's reason (e.g., "Adds a self-aware punchline to the famous quote.")
- **Post-accept** — the AI annotation tooltip shows only `AI Replacement` / model id / timestamp. The explanation is lost.

### The work

Plumb the explanation through accept:

1. At accept time, copy `suggestion.explanation` onto the annotation's metadata.
2. Render a third line in the tooltip (after model id / timestamp) showing the explanation when present.
3. If the model produced no explanation, omit the line — keep current tooltip layout.

No design call needed — the surface, the data, and the always-on behavior already exist. This is plumbing.

mrangelmarino:
@claude please implement this issue.

This is plumbing, not feature design — the issue body has the complete spec. Three steps:

1. At suggestion-accept time (find with `rg 'acceptSuggestion|accept.*[Ss]uggestion' src/renderer/extensions/ai-suggestions/`), read the suggestion's `explanation` field and copy it onto the annotation's metadata. Find the annotation extension that renders the post-accept tooltip — that's where the metadata needs to land.

2. Render a third line in the annotation tooltip, after the model ID and timestamp. Show the explanation when present; omit the line entirely when not present (keep current tooltip layout).

3. Don't change anything else — no design questions, no UI redesign, no behavior changes to the suggestion popover itself.

## Scope guardrails
- Touch only `src/renderer/extensions/ai-suggestions/` (accept handler) and the annotation extension's tooltip renderer.
- Do NOT touch `src/shared/tools/*` — another session is actively working there.
- Do NOT refactor the accept flow itself — just pipe the explanation through.

PR title: `feat(editor): preserve AI edit explanation in post-accept annotation tooltip (#447)`
PR body must include `Closes #447`.

PR #523, opened 2026-05-17
feat(#467): gate every mutating tool to Editor Mode (add_comment, resolve_comment, suggest_edit, accept_diff, reject_diff)

mrangelmarino:
## Summary

Chunk 4 of the #467 umbrella. Delivers the **non-authorship affordance** that #467 hinges on: in Editor Mode, the agent can flag judgment-bearing editorial concerns via `add_comment` without proposing the replacement prose. **Also tightens the dual-layer invariant codebase-wide** — every tool that mutates document or pending state now carries `requiresMode >= 'editor'`, so Chat Mode is read-only at the registry level, not just at the useChat call-site allowlist.

**Stacked on #522** — base branch is `issue-467-mode-rename`. Merge #522 first.

Refs #467, #468, #455 (which shipped the comment tools but didn't gate them for the built-in agent).

### `src/shared/tools/schemas/document.ts`

| Tool | Before | After | Why |
|---|---|---|---|
| `add_comment` | `requiresMode: null` | `requiresMode: 'editor'` | Mutates document state. Chat Mode is read-only; Editor needs it for the non-authorship affordance. |
| `resolve_comment` | `requiresMode: null` | `requiresMode: 'editor'` | Sibling of `add_comment` — removing a comment is also a mutation. |
| `list_comments` | `requiresMode: null` | **(unchanged)** | Read-only; stays available everywhere. |

### `src/shared/tools/schemas/editor.ts`

| Tool | Before | After | Why |
|---|---|---|---|
| `suggest_edit` | `requiresMode: null` | `requiresMode: 'editor'` | Creates a pending suggestion (document state mutation, even if not yet applied). Editor's defining capability. |
| `accept_diff` | `requiresMode: null` | `requiresMode: 'editor'` | Commits a pending suggestion into content. |
| `reject_diff` | `requiresMode: null` | `requiresMode: 'editor'` | Dismisses pending suggestion state. Symmetric with accept_diff. |

After this PR, every tool that mutates state has `requiresMode >= 'editor'`. The dual-layer invariant (registry gating + useChat allowlist) is true codebase-wide.

Combined with the `MODE_LEVEL` ladder from Chunk 3 (#522: `chat ⊂ editor ⊂ create`):

- **Chat Mode** → only read-only tools (`list_comments`, `read_document`, `get_outline`, etc.). Mutating comment tools and edit/suggest/accept/reject tools all excluded.
- **Editor Mode** → all the above plus `add_comment`, `resolve_comment`, `suggest_edit`, `accept_diff`, `reject_diff`.
- **Create Mode** → all the above plus `edit` and `insert` (direct writes).

### `src/renderer/hooks/useChat.ts`

Updated the comment on `CHAT_MODE_TOOL_NAMES` to reflect that the allowlist is now defense-in-depth alongside the registry-level mode gating. Both layers run; either one would catch a mutating tool leaking into Chat Mode, but having both means the failure mode is "tool missing" rather than "tool leaks" if anyone adds a future mutating tool with `requiresMode: null` by accident.

No behavioral change to the filter itself.

## Why both layers

Registry gating relies on every new tool author setting `requiresMode` correctly. The default for read-only tools is `null` — meaning "available in every mode" — so a brand-new mutating tool that forgets to set `requiresMode: 'editor'` would silently appear in Chat Mode. The explicit allowlist in `useChat.ts` keeps Chat Mode tight regardless: if someone adds `delete_paragraph` with `requiresMode: null`, it doesn't reach Chat Mode unless they also add it to `CHAT_MODE_TOOL_NAMES` deliberately.

## Prompt copy

The `suggest_edit` vs `add_comment` discrimination heuristic was added to `EDITOR_MODE_INSTRUCTIONS` back in Chunk 1 (#513). No prompt changes needed here.

## Verification

1. `npm run dev` — 142 modules transformed, main + preload + renderer build clean, app launches.
2. Switch to **chat** mode → ask the agent to list comments → works (`list_comments` still available). Ask the agent to add a comment → declines and redirects to Editor Mode.
3. Switch to **editor** mode → ask "the second paragraph has a competing thesis with the first" → agent uses `add_comment`, not `suggest_edit` (per the prompt discrimination heuristic from Chunk 1).
4. Switch to **create** mode → all comment tools still available, plus `edit` / `insert` for direct writes.

## Tests

No existing tests cover mode gating for these tools. A small unit-test suite is the natural insurance once vitest lands — assertions of the shape `isToolAvailableInMode('add_comment', 'chat') === false`. Worth a follow-up.

## Out of scope (in this chunk)

- Chunk 5 — accept-flow audit against #447's caption infrastructure (already completed in [#467 comment](https://github.com/solo-ist/prose/issues/467#issuecomment-4471340693); no PR required).
- `setAgentMode` dead-code removal — landed on #522 in commit [476ebdc](https://github.com/solo-ist/prose/pull/522/commits/476ebdc) in response to the same review pass.

### DIFF
--- a/src/renderer/components/chat/ChatInput.tsx
+++ b/src/renderer/components/chat/ChatInput.tsx
@@ -6,7 +6,7 @@ import { useEditorInstanceStore } from '../../stores/editorInstanceStore'
 import { useEditorStore } from '../../stores/editorStore'
 import { useChatStore, createMessageId } from '../../stores/chatStore'
 import { useCommandHistoryStore } from '../../stores/commandHistoryStore'
-import { executeTool, getAvailableTools } from '../../lib/tools'
+import { executeTool, getAvailableTools, isToolAvailableInMode } from '../../lib/tools'
 import { cn } from '../../lib/utils'
 import { useReviewStore } from '../../stores/reviewStore'
 
@@ -133,14 +133,16 @@ export function ChatInput({ onSend, isLoading, isStreaming, onStop }: ChatInputP
   // Disable input while app is initializing (loading draft/conversations) or while loading
   const isDisabled = isLoading || isInitializing
 
-  // All available commands (including built-in shortcuts)
-  // Normalize aliases: get_outline → outline for consistent display
+  // All available commands (including built-in shortcuts) filtered by the
+  // current ToolMode — don't suggest /suggest_edit in Chat Mode if it would
+  // fail at checkToolAccess. Normalize aliases: get_outline → outline.
   const allCommands = useMemo(() => [
     ...getAvailableTools()
       .filter(t => t !== 'create_and_open_file')
+      .filter(t => isToolAvailableInMode(t, toolMode))
       .map(t => t === 'get_outline' ? 'outline' : t),
     'help', 'clear', 'new', 'quick-review', 'review-diff'
-  ], [])
+  ], [toolMode])
 
   // Parse the current command from input (e.g., "/list_files " -> "list_files")
   const activeCommand = useMemo(() => {
@@ -217,7 +219,6 @@ export function ChatInput({ onSend, isLoading, isStreaming, onStop }: ChatInputP
     // Handle /help specially
     if (command.toolName === 'help') {
       const { addMessage } = useChatStore.getState()
-      const availableTools = getAvailableTools()
       addMessage({
         id: createMessageId(),
         role: 'user',
--- a/src/renderer/hooks/useChat.ts
+++ b/src/renderer/hooks/useChat.ts
@@ -38,7 +38,13 @@ const CHAT_MODE_TOOL_NAMES: ReadonlySet<string> = new Set([
 
 function getToolsForToolMode(toolMode: ToolMode): ReturnType<typeof getToolsForClaudeAPI> {
   if (toolMode === 'chat') {
-    return getToolsForClaudeAPI('create').filter((t) => CHAT_MODE_TOOL_NAMES.has(t.name))
+    // Pass 'chat' to the registry filter first so layer 1 actually fires for
+    // Chat Mode — only tools with `requiresMode <= chat` survive. Then narrow
+    // further to the explicit allowlist. Both layers run in series, so a
+    // regression in either (a missing `requiresMode: 'editor'` on a future
+    // mutating tool, or a missing entry in the allowlist) is caught by the
+    // other.
+    return getToolsForClaudeAPI('chat').filter((t) => CHAT_MODE_TOOL_NAMES.has(t.name))
   }
   return getToolsForClaudeAPI(toolMode)
 }
--- a/src/renderer/hooks/useChat.ts
+++ b/src/renderer/hooks/useChat.ts
@@ -15,13 +15,18 @@ import type { LLMMessage, LLMStreamToolCall, LLMContentBlock } from '../types'
 
 // Chat Mode gets a read-only tool subset rather than the full read+write
 // surface, so the agent can ground itself in the document without proposing
-// edits. We can't filter by `category === 'document'` alone — that category
-// includes the mutating add_comment and resolve_comment tools. Mutating
-// comment tools move behind Editor Mode via `requiresMode` in Chunk 4 of
-// #467; until then this explicit allowlist is the gate.
+// edits. Defense-in-depth alongside the registry-level mode gating:
+// `requiresMode` on each ToolConfig excludes mutating tools from Chat Mode at
+// the registry, AND this explicit allowlist ensures Chat Mode only ever
+// receives tools we have specifically vetted as read-only.
 //
-// Audit checkpoint: when new MCP/agent tools land (e.g., list_tabs, select_tab
-// from #450), decide whether they belong in Chat Mode and extend this set.
+// Why both layers: registry gating relies on every new tool author setting
+// `requiresMode` correctly; a tool added with `requiresMode: null` (the
+// default for read-only tools) would silently appear in Chat Mode. The
+// allowlist keeps the failure mode "tool missing" rather than "tool leaks."
+//
+// Audit checkpoint: when new tools land, decide whether they belong in
+// Chat Mode and extend this set.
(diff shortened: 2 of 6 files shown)


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/b1de1284-e113-4de6-aa1b-84f175c35f90?refs=aab7e56245583af620481a5f972b1169f4cc9413&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
Addressed observation #1 (defense-in-depth framing was partial) in [866de63](https://github.com/solo-ist/prose/pull/523/commits/866de63). All remaining mutating tools now have `requiresMode >= 'editor'`:

- `suggest_edit`: `null` → `'editor'`
- `accept_diff`: `null` → `'editor'`
- `reject_diff`: `null` → `'editor'`

After this, every tool that mutates document or pending state is gated at the registry level. The dual-layer invariant (registry gating + useChat allowlist) is now true codebase-wide rather than just for the comment tools.

No behavioral change in practice — the useChat allowlist already excluded these from Chat Mode. This makes the second layer real.

**On observation #2 (executeTool not consulting requiresMode):** worth flagging that this is actually already in place. `executeTool` at [src/renderer/lib/tools/index.ts:75](https://github.com/solo-ist/prose/blob/issue-467-editor-mode-add-comment/src/renderer/lib/tools/index.ts#L75) calls `checkToolAccess(toolName, mode)` and returns `MODE_RESTRICTED` if the tool isn't available in the supplied mode. A direct `executeTool('add_comment', …, 'chat', …)` returns an error, not a mutation. The mode gate is enforced at three points: (1) Anthropic API tool surface filtering in useChat, (2) the allowlist filter in useChat (defense-in-depth), and (3) the executor itself via `checkToolAccess`.

**Dismissed:** observation #3 (no test coverage) — fine given no vitest setup. The dual-layer invariant + executor gate are all type-checked, so the gating wouldn't drift silently.

**Rebase note:** branch rebased onto the updated #522 tip (which now includes the Shift+Tab fix); force-pushed with `--force-with-lease`. New commit count: 1 (was 1 before).

mrangelmarino:
/review 

mrangelmarino:
Addressed the body/scope mismatch by updating the PR description — `suggest_edit`, `accept_diff`, `reject_diff` are now in the "What changed" table where they belong, with the rationale for each. The "Out of scope" section is trimmed to only what's actually deferred (Chunk 5 audit, which is already closed without a PR per the umbrella plan).

You're right that this kind of doc/diff drift is exactly the thing future me grepping git log will get burned by. Fixed.

**Dismissed:**
- Aspirational `CHAT_MODE_TOOL_NAMES` comment — accurate as forward-looking guidance, just leans more on "future authors" now that every current mutating tool is gated. Reads fine.
- Test gap — agreed it's a real follow-up; tracking via this comment rather than filing a separate issue since the umbrella PR closes #467 anyway.

Branch rebased onto updated #522 tip; force-pushed with `--force-with-lease`.

mrangelmarino:
/review 

mrangelmarino:
Addressed the non-blocking nit in [1bb03db](https://github.com/solo-ist/prose/pull/523/commits/1bb03db) — switched `getToolsForClaudeAPI('create').filter(...)` → `getToolsForClaudeAPI('chat').filter(...)` so layer 1 (registry mode gate) actually fires for chat mode, then layer 2 (`CHAT_MODE_TOOL_NAMES` allowlist) narrows. Behavior unchanged today; the dual-layer invariant becomes load-bearing.

**Considered and backed off**: extending the registry gate to the file-write tools (`new_file`, `save_file`, `create_and_open_file`). They're `requiresMode: null` today, which means Chat Mode wouldn't have them via the agent surface (allowlist blocks) but **users invoking `/new` or `/new_file` slash commands** in Chat Mode would hit `MODE_RESTRICTED` because `ChatInput.tsx:472` passes the current `toolMode` to `executeTool`. That's a user-initiated path; gating it at the mode boundary is a real UX regression. Worth a separate scoped issue if we want to align user-initiated slash commands and agent-initiated tool calls under the same mode invariant.

Dismissed: aspirational allowlist comment (reads fine forward-looking) and test coverage (acknowledged follow-up).

mrangelmarino:
> **Auto-fix cycle exhausted.** A fix was already attempted for this PR. Escalating to human review.

mrangelmarino:
Tackled both non-blocking concerns in [aab7e56](https://github.com/solo-ist/prose/pull/523/commits/aab7e56):

**1. `checkToolAccess` error message** — was hardcoded `"Switch to Create Mode"` even when the tool only needs Editor. Now reads `tool.requiresMode` and labels the minimum-needed mode. A `suggest_edit` from Chat Mode now correctly says "Switch to Editor Mode."

**2. ChatInput slash-command autocomplete** — was listing every registered tool regardless of mode, so `/suggest_edit` and `/accept_diff` would appear as suggestions in Chat Mode and then fail at `checkToolAccess`. Filter added at the autocomplete construction (`ChatInput.tsx:139`) using `isToolAvailableInMode(t, toolMode)`. Exposed `isToolAvailableInMode` via `lib/tools/index.ts` re-export rather than reaching across to `shared/tools/registry` directly from the renderer.

**Also tidied** — dropped a dead `const availableTools = getAvailableTools()` from the `/help` handler (assigned but never used).

**Left as follow-up** — the hardcoded `/help` message body still lists every tool regardless of mode. Making that dynamic is a larger UX refactor (probably should pull tool descriptions from the registry too rather than hand-maintaining a markdown list). Worth a separate scoped issue.

**The submit-path** `getAvailableTools()` call (handleSubmit) intentionally stays mode-agnostic — it's a name-membership check for dispatching a typed slash command, and the mode-restriction error happens downstream at `executeTool` with the now-accurate message.

mrangelmarino:
/review 

Issue #533, opened 2026-05-18
Customize tool-call result rendering per tool (design pass)


mrangelmarino:
## Context

Caught during #467 QA. After unifying agent-driven and user-initiated tool calls to use the same \`<tool-result>\` tag format (so custom renderers work in both paths — see #533 or whatever this PR is once it lands), agent-driven tool results now show full JSON in chat for any tool that doesn't have a custom renderer. Consistent with what slash commands have always shown, but it's now a more visible UX surface because the agent invokes tools a lot more often than slash commands do.

We have a renderer registry at \`src/renderer/components/chat/toolResultRenderers/\` with two entries so far:

- \`list_files\` → reuses the file explorer's \`FileTree\` component
- \`request_mode_switch\` → a button-pair for one-click mode switching

Every other tool falls back to \`renderMarkdown\` of the JSON, which is unreadable for non-trivial payloads.

## What we want

Custom renderers (or compact summaries) for the remaining tools, picked per tool's actual UX value:

| Tool | Current | Ideas |
|---|---|---|
| \`read_document\` | Full JSON of node tree | Just \"Read N nodes\" + collapsible details |
| \`read_selection\` | JSON with text + positions | Quote-styled block with the selected text |
| \`get_metadata\` | Flat key/value JSON | Two-column key/value table or dl-list |
| \`get_outline\` | JSON array of headings | Render as actual nested outline, clickable to jump-to-line |
| \`search_document\` | JSON array of matches | List with line refs, click to jump |
| \`list_comments\` | JSON array of comments | Comment cards with marked text + body, mirror the editor's comment popover style |
| \`add_comment\` | JSON with new comment id | Single-line confirmation \"Added comment to '<marked text>'\" with a click-to-locate |
| \`resolve_comment\` | JSON with success | Single-line \"Resolved comment\" |
| \`suggest_edit\` | JSON with suggestionId | Already partially custom (clickable suggestion); polish further |
| \`accept_diff\` / \`reject_diff\` | JSON | Single-line confirmations |
| \`edit\` / \`insert\` | JSON | Single-line confirmation with the affected range |
| \`open_file\` | JSON | \"Opened: <basename>\" with click to focus that tab |
| \`new_file\` | JSON | \"Created and opened: <basename>\" |
| \`save_file\` | JSON | \"Saved to <path>\" |
| \`read_file\` | JSON with content | Snippet preview + collapse-to-expand |
[...]

PR #532, opened 2026-05-18
feat(chat): mid-conversation mode-switch awareness via system prompt

mrangelmarino:
## Summary

Caught during the #467 manual QA pass. When a user switches modes mid-conversation (e.g., starts in Editor, switches to Chat after a few turns), the agent's behavior can lag because:

- The system prompt rebuilds per turn with current-mode instructions ✓
- But the conversation history reflects the *previous* mode's dynamics — prior tool calls, prior posture
- The agent pattern-matches on history and can drift back toward the old mode's framing

## Fix

Track \`lastSentToolMode\` in \`chatStore\`. On every send, if \`toolMode !== lastSentToolMode && lastSentToolMode !== null\`, prepend a one-line note to the system prompt:

> Note: the user just switched to <Mode> Mode mid-conversation. Earlier turns may have been in a different mode with different capabilities. Adopt the current mode's posture immediately.

**Idempotent.** Only the diff at SEND time matters. Multiple toggles between sends collapse to a single marker for the final mode. \`lastSentToolMode\` updates immediately, so the tool-loop continuation within the same stream doesn't re-fire the marker on every roundtrip.

**Invisible to chat UI.** The marker lives only in the system prompt sent to the API. No bubble clutter, no rendering changes.

## Files

- \`src/renderer/stores/chatStore.ts\` — \`lastSentToolMode\` field + setter, initialized \`null\`, in-memory only (matches \`toolMode\` itself)
- \`src/renderer/lib/prompts.ts\` — \`buildSystemPrompt\` takes optional \`modeJustSwitched\`, prepends the note when true
- \`src/renderer/hooks/useChat.ts\` — both send paths (initial + tool-loop continuation) compute the flag, update \`lastSentToolMode\` atomically with send

## Verification

1. In Editor mode, send a message. \`lastSentToolMode\` becomes \`editor\`.
2. Switch through Chat → Create → Chat (4 toggles, no sends).
3. Send a message in Chat. The API request's \`system\` field has \`Note: the user just switched to Chat Mode...\` prepended. Only one marker, for the final mode.
4. Send another message in Chat without switching. **No marker** (mode unchanged since last send).
5. Switch to Editor, send a message → marker fires (\`switched to Editor\`).
6. Toggle mode mid-stream during a tool-loop continuation → marker also fires in the loop's next continuation request (rare but supported).

[...]

PR #534, opened 2026-05-18
feat(chat): request_mode_switch tool + persona softening

mrangelmarino:
## Summary

Caught during the #467 QA pass. The agent's refusal language when a request was out of scope for the current mode felt didactic. The user had to read the explanation, click the StatusBar dropdown to switch modes, then re-type their prompt — five steps for what should be one click. Example from the QA log:

> User: \"Write me a paragraph about coffee.\"
> Agent: \"That's authorship — you'd be putting my prose into your document under your name. Not something I'll do by default. If you want to draft it yourself, I'm happy to react: tell me the angle (sensory? historical? a complaint about a specific café?), give me a rough first pass, and I'll push back on what's working and what isn't. If you genuinely want LLM-authored prose, switch to Create Mode and re-ask — that's the mode where you've opted in.\"

This PR adds a new tool the agent can call instead of lecturing, plus prompt updates so it actually uses it.

## New tool: \`request_mode_switch\`

Schema in \`src/shared/tools/schemas/ui.ts\`:

\`\`\`ts
{
  target: 'chat' | 'editor' | 'create',
  reason: string,           // one-sentence what-this-enables
  prompt_to_retry: string   // the exact prompt the agent would run after the switch
}
\`\`\`

- \`requiresMode: null\` — available in every mode; it's the agent's escape valve
- \`category: 'ui'\`
- Added to \`CHAT_MODE_TOOL_NAMES\` allowlist so the read-only Chat Mode also gets it
- Executor is a no-op pass-through: never changes the mode itself, just packages the agent's request

## Renderer

\`src/renderer/components/chat/toolResultRenderers/RequestModeSwitchResult.tsx\` renders the tool result as a small inline affordance:

\`\`\`
[reason]
\"[prompt_to_retry]\"
[Switch & Run] [Just Switch]
\`\`\`

- **Switch & Run** → \`setToolMode(target)\` + \`sendMessage(prompt_to_retry)\`. One click, agent gets to work in the new mode.
- **Just Switch** → \`setToolMode(target)\` only. User can edit or re-ask manually.

\`sendMessage\` is dispatched through a \`CustomEvent\` (\`MODE_SWITCH_RUN_EVENT\`) so the renderer doesn't have to be a \`useChat\` consumer — \`ChatPanel\` listens for the event and calls its local \`sendMessage\`. Renderer disables both buttons after first click to prevent double-fire.

## Unified tool-result format

Agent-driven tool calls were producing inline summaries (\`*[Tool: foo] Success*\`) that custom renderers couldn't parse — they got a literal string \"Success\" instead of structured data. Caught in QA when the mode-switch renderer kept showing \"Unable to parse mode-switch request.\"

Switched the agent-path message format in \`useChat.ts\` to use the same \`<tool-result name=\"...\" success=\"...\">JSON</tool-result>\` tag format that user-initiated slash commands use. Now custom renderers get the full structured payload regardless of who triggered the tool call. Tools without a custom renderer fall back to the markdown render of the JSON — same as slash commands have always shown. Per-tool render polish tracked in #533.

## Persona softening

\`src/renderer/lib/prompts.ts\`:

- **\`BASE_PROMPT\` authorship section** — replaced the \"stop and ask whether the user wants you to leave the drafting to them\" line with a directive to call \`request_mode_switch\` and not lecture about modes
- **\`CHAT_MODE_INSTRUCTIONS\`** — new \"When a request needs a different mode\" section pointing at the tool, replacing the previous \"tell the user to switch to Editor Mode (StatusBar → editor)\" prose
- **\`EDITOR_MODE_INSTRUCTIONS\`** — drafting requests now route through the tool instead of a prose decline

Effect: refusals become a short prose sentence + a tool call rendering a button. No more multi-paragraph lectures about authorship.

## Files

- New: \`src/shared/tools/schemas/ui.ts\` — \`requestModeSwitchConfig\`
- New: \`src/renderer/lib/tools/executors/ui.ts\` — pass-through executor
- New: \`src/renderer/components/chat/toolResultRenderers/RequestModeSwitchResult.tsx\` — button-pair UI
- \`src/shared/tools/registry.ts\` — register \`uiTools\`
- \`src/renderer/lib/tools/index.ts\` — dispatch \`request_mode_switch\`
- \`src/renderer/components/chat/toolResultRenderers/index.tsx\` — registry entry
- \`src/renderer/components/chat/ChatPanel.tsx\` — \`MODE_SWITCH_RUN_EVENT\` listener
- \`src/renderer/hooks/useChat.ts\` — allowlist + unified tool-result format
- \`src/renderer/lib/prompts.ts\` — persona softening + tool teaching in 3 instruction blocks

## Verification

1. Chat Mode, new conversation. Open a doc with a typo. Send \"Fix the typo in paragraph 2\" → agent emits \`request_mode_switch\` with \`target: 'editor'\`. UI renders reason + boxed prompt + two buttons.
2. Click **Switch & Run** → mode flips to Editor, agent immediately re-runs the prompt, uses \`suggest_edit\`.
3. Repeat in Chat Mode. Send \"Write me a paragraph about coffee\" → agent emits with \`target: 'create'\`.
4. Click **Just Switch** → mode flips to Create, no auto-send.
5. Editor Mode, drafting request → same path, target Create.
6. Verify the agent no longer produces multi-paragraph refusal lectures.

## Follow-ups

- **#533** — per-tool result renderer design pass. The unified tool-result format exposes JSON for tools without a custom renderer; worth polishing case-by-case.

### DIFF
--- a/docs/architecture/llm-pipeline.md
+++ b/docs/architecture/llm-pipeline.md
@@ -49,7 +49,7 @@ const MAX_HISTORY_MESSAGES = 20   // context window pruning
 Two refs live outside the React hook so globally-registered IPC event handlers can access them:
 
 - **`pendingToolCallsRef`** — accumulates `{id, name, args}` from streaming `tool-call` events, tagged with `streamId`
-- **`toolLoopContextRef`** — tracks `{apiMessages, assistantMsgId, roundtripCount, lastErrorSignature}` across recursive stream restarts, tagged with `streamId`
+- **`toolLoopContextRef`** — tracks `{apiMessages, assistantMsgId, roundtripCount, lastErrorSignature, modeJustSwitched}` across recursive stream restarts, tagged with `streamId`. `modeJustSwitched` is set at turn start and preserved across continuations so the "you just switched modes" notice stays on the system prompt for every LLM call within a turn that began with a mode switch.
 
 Both are reset by `clearStreamRefs()` on abort, error, completion, or circuit-breaker stop.
 
--- a/src/renderer/components/chat/ChatMessage.tsx
+++ b/src/renderer/components/chat/ChatMessage.tsx
@@ -1,7 +1,7 @@
 import { useState, useCallback, useEffect, useRef, useMemo } from 'react'
 import { cn } from '../../lib/utils'
 import type { ChatMessage as ChatMessageType } from '../../types'
-import { User, Wand2, Check, AlertCircle, ArrowRight, Sparkles, CheckCircle2, XCircle, RotateCcw } from 'lucide-react'
+import { User, Wand2, Check, AlertCircle, ArrowRight, Sparkles, CheckCircle2, XCircle, RotateCcw, ChevronDown } from 'lucide-react'
 import { parseEditBlocks, hasEditBlocks, stripEditBlocks, type EditBlock } from '../../lib/editBlocks'
 import { applyEditsAsDiffs, applyEditsDirect, type ApplyResult, type EditProvenance } from '../../lib/applyEdit'
 import { useEditorInstanceStore } from '../../stores/editorInstanceStore'
@@ -16,7 +16,16 @@ import avatarDark from '../../assets/avatar-dark.png'
 import avatarLight from '../../assets/avatar-light.png'
 
 // Tool call indicator component with AI sparkle styling
-function ToolCallIndicator({ name, status, onClick, children }: { name: string; status: 'executing' | 'success' | 'error'; onClick?: () => void; children?: React.ReactNode }) {
+function ToolCallIndicator({ name, status, onClick, children, actions, defaultExpanded = false }: { name: string; status: 'executing' | 'success' | 'error'; onClick?: () => void; children?: React.ReactNode; actions?: React.ReactNode; defaultExpanded?: boolean }) {
+  const [expanded, setExpanded] = useState(defaultExpanded)
+  const hasBody = children != null && status !== 'executing'
+  const hasActions = actions != null
+
+  const toggleExpanded = (e: React.MouseEvent) => {
+    e.stopPropagation()
+    setExpanded(v => !v)
+  }
+
   return (
     <div
       className={cn(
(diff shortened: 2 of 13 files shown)


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/62b0b955-eef8-41b9-941a-a2067995b83a?refs=398e166112f6de82225bb32b4c719dbf2c5ac06e&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
Pushed [9f23339](https://github.com/solo-ist/prose/pull/534/commits/9f23339) addressing a double-fire bug caught in QA:

**Sequence that broke:**
1. Chat Mode, "fix the typo" → agent emits \`request_mode_switch\` ✓
2. Click Switch & Run → \`setToolMode('editor')\` + dispatchEvent in same tick
3. \`sendMessage(prompt_to_retry)\` runs inside the listener handler — but the captured \`toolMode\` was still \`'chat'\` because React hadn't re-rendered
4. Turn 2 hits the API with **Chat Mode system prompt** → agent emits \`request_mode_switch\` again

**Fix 1 — stale closure:** \`sendMessage\` now reads \`toolMode\` fresh from \`useChatStore.getState()\` inside its body instead of relying on the React-captured value from the \`useCallback\` closure. Always reflects the latest store state regardless of render timing.

**Fix 2 — LLM pattern-match defense in depth:** even with the correct Editor Mode system prompt for turn 2, the agent might still reflexively re-emit \`request_mode_switch\` from history pattern-matching. Tightened two prompt surfaces:

- The mode-switch-awareness marker (added in #532) now reads "**Do NOT call request_mode_switch in this turn** — they've already switched"
- The \`request_mode_switch\` tool description has a "DO NOT call this tool if you can fulfill the user's request with the tools already available" guard

Both fixes confirmed in QA after Cmd+R: turn 2 goes straight to \`read_document\` + \`suggest_edit\`, no spurious mode switch.

mrangelmarino:
/review 

PR #554, opened 2026-05-19
fix(editor): chunked streaming appends inline text, not new paragraphs (#553)

mrangelmarino:
## Summary

Follow-up fix to **#544** (cosmetic chunked streaming for agent `insert`/`edit`, originally tracked under #541). #544's implementation chunks the streaming correctly but each chunk lands as its own paragraph block, defeating the "word-by-word fill in one paragraph" design intent.

## Root cause

`applyInsertion` (added by #544) called `editor.chain().insertContent(chunk)` for every streaming chunk past the first. `insertContent` with a plain string routes through the configured markdown content parser, which wraps top-level text in a paragraph node — so each chunk landed as its own paragraph block instead of appending inline to the paragraph the first chunk just created.

Visible symptom (reproduced 2026-05-19 during #467 Stage 6.1 QA): a ~50-word "Write a paragraph about coffee" response came in as 18 paragraph blocks (3 words each, since `STREAM_TARGET_FRAME_COUNT = 24` produces `Math.ceil(50/24) = 3` words per chunk) rather than one paragraph filling word-by-word. The streaming animation also wasn't perceptible because the paragraph-block inserts complete in roughly one paint cycle.

## Change

Single file: `src/renderer/lib/tools/executors/editor.ts`. +8 / -1.

Subsequent chunks now use a raw ProseMirror `tr.insertText` dispatch instead of `editor.chain().insertContent(chunk)`. `tr.insertText` inserts inline text into the surrounding block, preserving paragraph continuity and letting the rAF-gated chunked apply produce a visible word-by-word fill within a single paragraph.

The first chunk still goes through `insertContent` so range-replacement (`insert(position='cursor')` over a selection) and block-level setup behave the same as in #544.

## Scope and limitations

Targets the common single-paragraph drafting case. For multi-paragraph drafting where the agent's text contains `\n\n`, chunks containing newlines would now insert as literal characters (paragraph breaks lost during streaming) instead of as separate paragraphs. Best-effort; tracked separately if it bites further QA.

## Test plan

- [ ] Create Mode, "Write me a paragraph about coffee." → paragraph fills word-by-word **in a single paragraph block** over ~400ms.
- [ ] System Settings → Accessibility → Display → Reduce Motion enabled → paragraph appears instantly in one paragraph block (existing instant-apply path).
- [ ] Settings → Editor → "Stream agent edits" off → paragraph appears instantly in one block.
[...]

Issue #541, opened 2026-05-19
Stream LLM-generated insertions and deletions into the document body


mrangelmarino:
## Motivation

`executeInsert` and `executeEdit` apply their content via a single `editor.chain().focus().insertContent(text).run()`. From the user's perspective the paragraph just *appears* — they can't read along, react, or stop mid-write. This is fine for one-line typo fixes; for multi-sentence drafts in Create mode it makes the agent feel like a printer rather than a collaborator.

For agentic editing to feel alive, the document should show content arriving in chunks the user can perceive — at a minimum paragraph-by-paragraph, ideally word-by-word.

### A) Cosmetic streaming (recommended first cut)

Once the tool call arrives, split the text into chunks (word or sentence boundary) and insert each chunk in a `requestAnimationFrame` loop. Total wall time bounded (~200–500ms for a paragraph); user sees motion.

- Pros: localized change in `executeInsert` / `executeEdit`. No protocol changes.
- Cons: text was already fully generated by the LLM before any of it shows up — *"cancel mid-write"* isn't possible.

### B) True streaming via partial tool-args

Anthropic streaming exposes `input_json_delta` events during tool-use generation. The renderer pipeline could parse partial JSON and incrementally apply the text as it streams in.

- Pros: real-time write; user can intervene before the LLM finishes generating.
- Cons: cross-cutting change — main-process tool dispatch (`src/main/ipc.ts` LLM stream handler), shared tool-call protocol, executor signatures. Probably needs a feature-flag rollout.

**Default to Approach A** for this issue. File B as a follow-up if A lands and we want the further win.

## Files (Approach A)

- `src/renderer/lib/tools/executors/editor.ts` — chunked insertion loop in `executeInsert` / `executeEdit`
- `src/renderer/types/index.ts` and `src/main/ipc.ts` defaults — `streamingEdits: boolean` setting (default on; off for users who prefer instant apply)
- `src/renderer/components/settings/` — surface the toggle in Editor settings

## Acceptance criteria

1. Inserting a multi-sentence paragraph via the agent in Create mode shows the text arriving in visible chunks rather than appearing instantly.
2. Streaming respects `prefers-reduced-motion` — when set, fall back to instant apply.
3. Settings toggle (default on) disables the effect for power users.
4. No regression in undo behavior — the entire streamed insertion still groups into a single undo step.
5. Annotation provenance (start/end positions) remains correct after streaming completes.

## Out of scope

[...]

PR #555, opened 2026-05-19
fix(editor): track chunked-stream position via selection, not doc-size delta

mrangelmarino:
## Summary

Follow-up to **#554** (and ultimately #544's chunked streaming for `insert` / `edit`). #554 correctly switched subsequent chunks from `insertContent(chunk)` to a raw `tr.insertText`, but the running position math wasn't updated to match — chunks still stacked as one-paragraph-per-chunk.

## Symptom (still present after #554 merged)

Reproduced 2026-05-19 in Create Mode against the running dev on `release/v1.2-agent-persona @ 06bff28`:

> The chunks now arrive progressively over ~400ms (the rAF-gated streaming animation IS visible), but **each chunk still lands as its own paragraph block** — visually stacked, one word/group per line, blank-line spacing between.

## Root cause

`applyInsertion` was computing the running insertion endpoint as `pos = from + (doc.content.size - sizeBefore) + rangeSize`. When the first chunk's `insertContent` creates a new paragraph block, the doc-size delta includes the paragraph's opening + closing tokens (2 PM units of overhead). So `pos` lands at the position **after** the closing paragraph token — a between-blocks position.

`tr.insertText` at a between-blocks position calls `replaceRangeWith`, which wraps the inserted text in a new paragraph to fit the parent context (per ProseMirror's smart-replace behavior). Result: every chunk creates its own paragraph despite using the inline-text primitive.

## Fix

Anchor the running endpoint at end-of-insertion using TipTap's own cursor placement (`selectionToInsertionEnd` in `node_modules/@tiptap/core/src/commands/insertContentAt.ts:179`), which leaves the cursor INSIDE the just-inserted textblock at the end of its text. From there, advance `pos` via an explicit length accumulator (`pos += chunks[i].length`) rather than re-reading `editor.state.selection.from` after every dispatch.

Two reasons for the accumulator over a live-selection read:

1. Subsequent chunks use raw `tr.insertText` (not `insertContent(string)`), which inserts plain text inline and adds exactly `chunks[i].length` PM units — so the accumulator is exact.
2. If the user clicks or types mid-stream, the live selection moves away from our insertion endpoint. The accumulator stays independent, keeping later chunks landing at the running insertion point instead of redirecting into the user's cursor — preserving the invariant the original doc-size approach was designed to protect.

Function-level docstring updated to reflect the new tracking strategy. First-chunk path is unchanged. Single file: `src/renderer/lib/tools/executors/editor.ts`. +25 / -17.

## Verified locally

Tested live via HMR in Create Mode before pushing:
- "Write me a paragraph about coffee" → paragraph fills word-by-word inside a single paragraph block over ~400ms. No stacking.
- Multiple subsequent insert / edit / suggest_edit / delete / move_cursor cycles exercised in the same session — all behaved as expected.

## Test plan

- [ ] Create Mode, "Write me a paragraph about coffee." → paragraph fills word-by-word inside a single paragraph block over ~400ms.
- [ ] With reduced motion or `streamingEdits: false`: paragraph appears instantly in one block (instant-apply path unchanged).
- [ ] Highlight a sentence, "use insert to replace this with: '…'" → replaced span is a single paragraph.
- [ ] Mid-stream: click elsewhere in the document during a streamed insert — remaining chunks continue landing at the original insertion endpoint, not at the click position.

Refs #467, #541, #544, #554

### DIFF
--- a/src/renderer/lib/tools/executors/editor.ts
+++ b/src/renderer/lib/tools/executors/editor.ts
@@ -87,9 +87,11 @@ function nextFrame(): Promise<void> {
  * the same chain (a single transaction), so range replacement is atomic and
  * shares undo state with the first chunk.
  *
- * Subsequent chunks track the running insertion endpoint explicitly via the
- * doc-size delta, so user clicks or keystrokes mid-stream can't redirect
- * later chunks into the user's cursor position.
+ * Subsequent chunks track the running insertion endpoint via an explicit
+ * length accumulator (anchored once at end-of-insertion from TipTap's own
+ * selection placement, then advanced by each chunk's character count), so
+ * user clicks or keystrokes mid-stream can't redirect later chunks into
+ * the user's cursor position.
  */
 async function applyInsertion(
   editor: Editor,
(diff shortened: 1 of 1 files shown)


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/be56b993-d1b8-4120-976f-3933b0132b73?refs=d90c1e5a77488a82afac4b9a0406232644ee3a76&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
/review 

PR #590, opened 2026-05-25
fix(updater): apply staged updates on quit (#577)

### LINKED ISSUE #577: Auto-update downloads but doesn't apply on restart (macOS) — user stays on old version
## Summary

Auto-update **downloads successfully but never applies on restart** — the app relaunches (or stays running) on the old version, with the new version staged-but-unapplied in the updater cache.

Reproduced live by inspecting a running installed **v1.2.1** with **v1.3.0** published.

## Evidence (from the running install)

- **Installed / running:** `/Applications/Prose.app` — `CFBundleShortVersionString = 1.2.1`.
- **Feed is complete:** v1.3.0 release has `latest-mac.yml`, `Prose-1.3.0-arm64-mac.zip` (+ blockmap), and the DMG. Nothing missing on the publish side.
- **Update WAS downloaded and staged** in `~/Library/Caches/prose-updater/`:
  - `update.zip` (151 MB), `current.blockmap`
  - `pending/Prose-1.3.0-arm64-mac.zip`
  - `pending/update-info.json` → `{ "fileName": "Prose-1.3.0-arm64-mac.zip", "sha512": "LKeath3K9...Di0Lw==", "isAdminRightsRequired": false }`
  - Staged today; app still reports **1.2.1**.

So the download + staging path works; the **quitAndInstall / relaunch step is what fails** — the staged `pending/` update is never applied.

## Likely root cause (`src/main/updater.ts`)

`autoInstallOnAppQuit = true`, but the macOS `quitAndInstall` hardening lives **only** in the explicit `updater:install` IPC handler:

- `markQuitting()` — prevents the macOS hide-on-close handler from swallowing the window close (otherwise the app hides in the background instead of quitting+relaunching, "window vanishes, no relaunch").
- `installInvoked` re-entrancy guard — a second `quitAndInstall` double-registers a native observer and trips a `NOTREACHED` (`electron_api_auto_updater.cc:118`) that bails **before** the quit/relaunch (referenced in-code as **Sentry PROSE-H**).

The **`autoInstallOnAppQuit` path** (fired by electron-updater on a normal app quit) bypasses both guards. On macOS the hide-on-close handler likely swallows the close → no quit → staged update never applied. The two code paths (auto-on-quit vs. explicit install button) are not protected equally.

## Steps to reproduce

1. Install an older release (e.g. 1.2.1) in `/Applications`.
2. Publish a newer release (1.3.0).
3. Let the app detect + download the update (stages into `prose-updater/pending/`).
4. Quit / restart the app (or use the in-app "Restart to update").
5. **Expected:** relaunches on 1.3.0. **Actual:** still 1.2.1; `pending/` update remains.

## Investigation leads

1. Route the `autoInstallOnAppQuit` path through the same `markQuitting()` + re-entrancy guard as `updater:install` — or set `autoInstallOnAppQuit = false` and install **only** via the guarded handler.
2. Audit the macOS hide-on-close / `before-quit` interaction with `quitAndInstall` (does `markQuitting()` actually let the windows close + relaunch fire?).
3. Check **Sentry PROSE-H** for field occurrences of the `NOTREACHED` bail.
4. **No persistent updater logging** — no `electron-log` file was found on disk (updater logs go to console only). Wiring `autoUpdater.logger` to a file would make this diagnosable in the field. (Secondary finding.)

## Impact

Existing users won't auto-update to **v1.3.0** on restart — they'll need a manual DMG download. Pre-existing updater logic (not introduced by 1.3.0), but it gates the 1.3.0 rollout.

mrangelmarino:
## Summary

Fixes the macOS auto-update bug where a new version **downloads and stages but never applies on restart** — the app relaunches on the old version with the staged update left in `~/Library/Caches/prose-updater/pending/`.

Closes #577

## Root cause

`autoInstallOnAppQuit = true` made electron-updater apply the staged update from the **`quit` event**. On macOS that's too late — the app is already terminating, so Squirrel.Mac never stages the relaunch. That path also bypassed both guards that only the explicit `updater:install` handler had:

- `markQuitting()` — without it, the macOS hide-on-close handler swallows the window close (app hides instead of quitting+relaunching).
- `installInvoked` re-entrancy guard — a second `quitAndInstall` trips a native `NOTREACHED` (`electron_api_auto_updater.cc:118`) that bails before quit/relaunch (Sentry PROSE-H).

So the two install paths (auto-on-quit vs. explicit button) were not protected equally, and the auto path fired at the wrong moment.

## Changes (`src/main/updater.ts`)

- **Disable `autoInstallOnAppQuit`** and apply staged updates ourselves.
- **Single guarded `installUpdate()`** — both the "Restart" button (`updater:install`) and the auto-on-quit path funnel through it, so `markQuitting()` + the re-entrancy guard apply uniformly.
- **Trigger auto-install from `before-quit`** (not `quit`) with `event.preventDefault()`, so `quitAndInstall` drives a clean quit+relaunch early enough for Squirrel — rather than racing a half-finished termination.
- **Persistent updater log** — wire `autoUpdater.logger` to a file under the OS logs dir (`~/Library/Logs/Prose/updater.log`), since updater output previously went to console only and was lost on packaged builds (investigation lead #4). Logging is fire-and-forget and degrades to console-only if the dir isn't writable.

## Verification

This is **dev-disabled** (`if (is.dev) return`) and only exercisable in a packaged build against a published newer release, so it can't be verified via the dev server / HMR:

- ✅ `npm run build` compiles cleanly (electron-vite main + preload + renderer).
- ✅ No new TypeScript errors in `updater.ts`.
- ⏳ **End-to-end (requires release pipeline):** install an older signed build in `/Applications`, publish a newer release, let it download, then **Cmd+Q** (or click "Restart"). Expected: relaunches on the new version; `prose-updater/pending/` is consumed. Check `~/Library/Logs/Prose/updater.log` for `Applying staged update via quitAndInstall`.

### DIFF
--- a/src/main/updater.ts
+++ b/src/main/updater.ts
@@ -92,10 +92,17 @@ export async function initAutoUpdater(mainWindow: BrowserWindow): Promise<void> 
 
     let updateDownloaded = false
     let installInvoked = false
+    // Set if quitAndInstall() throws before it can quit/relaunch. Gates the
+    // before-quit auto-apply so a persistently failing install can't trap the
+    // user in a can't-quit loop — the staged update simply retries on next
+    // launch instead.
+    let installFailed = false
 
     // Single guarded entry point for applying a staged update. Both the explicit
     // install button (updater:install) and the auto-install-on-quit path funnel
     // through here so the macOS quit/relaunch hardening is applied uniformly.
+    // Returns true once quitAndInstall has been handed off, false if it could
+    // not start (no update, re-entrant call, or a thrown failure).
     const installUpdate = (): boolean => {
       if (!updateDownloaded) return false
       // Re-entrancy guard: a second QuitAndInstall double-registers an observer
@@ -114,8 +121,19 @@ export async function initAutoUpdater(mainWindow: BrowserWindow): Promise<void> 
       // but no relaunch.
       markQuitting()
       logger.info('Applying staged update via quitAndInstall')
-      autoUpdater.quitAndInstall()
-      return true
+      try {
+        autoUpdater.quitAndInstall()
+        return true
+      } catch (err) {
+        // quitAndInstall threw before it could quit/relaunch. Re-arm so an
+        // explicit "Restart" can retry this session, and flag the failure so the
+        // before-quit handler stops intercepting (the user must still be able to
+        // quit). The staged update stays in pending/ and retries on next launch.
+        installInvoked = false
+        installFailed = true
+        logger.error('quitAndInstall failed:', err instanceof Error ? err.message : String(err))
+        return false
+      }
     }
 
     autoUpdater.on('update-available', (info) => {
@@ -161,9 +179,13 @@ export async function initAutoUpdater(mainWindow: BrowserWindow): Promise<void> 
     // quitAndInstall can drive a clean quit+relaunch itself rather than racing a
     // half-finished termination.
     app.on('before-quit', (event) => {
-      if (!updateDownloaded || installInvoked) return
+      if (!updateDownloaded || installInvoked || installFailed) return
       event.preventDefault()
-      installUpdate()
+      // If quitAndInstall can't even start, honor the user's quit so they're
+      // never left unable to exit; the update retries on next launch.
+      if (!installUpdate()) {
+        app.quit()
+      }
     })
 
     // IPC handlers
--- a/src/main/updater.ts
+++ b/src/main/updater.ts
@@ -1,7 +1,58 @@
-import { BrowserWindow, ipcMain } from 'electron'
+import { app, BrowserWindow, ipcMain } from 'electron'
 import { is } from '@electron-toolkit/utils'
+import { mkdirSync } from 'node:fs'
+import { appendFile } from 'node:fs/promises'
+import { dirname, join } from 'node:path'
 import { IS_MAS_BUILD } from './env'
 import { markQuitting } from './quitState'
+
+// Persistent updater logger. electron-updater accepts any object with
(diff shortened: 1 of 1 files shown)


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/d671fa05-4990-405d-bf8d-e4045baf7485?refs=b31929a6d97baafff31ef4d82078a4d21a0640ae&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
Addressed the review's `installInvoked`-reset suggestion in b31929a.

I diverged slightly from the literal suggestion because resetting `installInvoked` **alone** would let the `before-quit` handler call `preventDefault()` + retry on *every* quit — if `quitAndInstall()` kept failing, the user would be **trapped, unable to exit the app**. That's a worse failure than the original.

So instead:
- `quitAndInstall()` is wrapped in try/catch.
- On failure: reset `installInvoked` (so an explicit **Restart** button press can still retry this session) **and** set an `installFailed` flag that makes the `before-quit` handler stop intercepting.
- The `before-quit` handler now calls `app.quit()` when the install can't start, so the user's quit is **always honored**. The staged update stays in `pending/` and retries on next launch.

Net: a failed `quitAndInstall()` degrades to "update applies next launch" instead of "user can't quit." Build is green locally.

mrangelmarino:
/review 

Issue #578, opened 2026-05-24
suggest_edit clobbers the whole document body when it's a single paragraph node (data loss)


mrangelmarino:
## Summary

A routine `suggest_edit` intended to fix a typo in the **first line** silently **deleted the entire rest of the document body**. Recovery was only possible because the full text was still in the chat model's context — had the conversation been cleared, the content would have been lost.

Surfaced during the #511 (appearance foundation) qa-hitl pass while exercising the agent. **Unrelated to #511** — this is in the `suggest_edit` / document node-structure path (shipped in v1.3.0), not the appearance work. Filing as a deferred follow-up.

## What happened

1. A short multi-paragraph piece ("The Last Cartographer") was open. The first line had a typo (`stoppedbeing`, missing space).
2. User asked the agent to "suggest some improvements." The agent called `read_document`, then `suggest_edit` to fix the first-line typo.
3. **Accepting the edit replaced the entire document body with just the corrected first sentence** — every other paragraph was gone. `read_document` afterward returned only the heading + the single fixed sentence.
4. The agent restored the content via another `suggest_edit`.
5. **After restore, the whole body came back as a single paragraph node** rather than discrete paragraphs.

## Evidence (live editor state, captured via CDP during the session)

```json
{
  "blockCount": 2,
  "blocks": [
    { "tag": "h2", "id": "j2dkj4ue", "textLen": 21,   "preview": "The Last Cartographer" },
    { "tag": "p",  "id": "o44y47hx", "textLen": 1818, "preview": "The maps stopped being accurate around the ti" }
  ]
}
```

The entire ~1,800-character body is **one `<p>` node** under the heading.

## Likely root cause

The document body is stored/rendered as a **single paragraph node** instead of being split into paragraph nodes (paragraph breaks appear to be soft line breaks within one node). Given a single-node body:

- A node-targeted `suggest_edit` ("fix the typo in the first line") operates on the *entire* 1,800-char node. The model emits only the corrected first sentence as the node's new content → **everything else in the node is overwritten** (the data loss in step 3).
- Node-targeted suggestions can't address individual paragraphs at all (the rendering quirk the user noticed in step 5).

So the single-node structure is both the **cause of the data loss** and a **standalone editing limitation**. Fixing the node splitting likely addresses both; a defensive guard so `suggest_edit` never silently drops content from a node it's only partially editing would be worthwhile regardless.

## Expected

- Multi-paragraph content should parse into multiple paragraph nodes (each with its own node id).
[...]

PR #619, opened 2026-05-26
fix(editor): prevent suggest_edit data-loss on single-paragraph docs (#578, #516)

### LINKED ISSUE #516: fix(editor): tighten suggest_edit frontmatter-only fall-through to require nodeId='frontmatter'
## Problem

When \`suggest_edit\` is called with content that's frontmatter-only (body empty after stripping \`---\`-wrapped YAML), the executor stages the frontmatter overlay and returns success regardless of what \`nodeId\` was passed:

\`\`\`ts
// src/renderer/lib/tools/executors/editor.ts (around the staged-then-commit block)
if (frontmatterToStage && !content.trim()) {
  useEditorStore.getState().setPendingFrontmatter(frontmatterToStage)
  return toolSuccess({ suggested: true, suggestionId: generateId() })
}
\`\`\`

This means a call like \`suggest_edit({ nodeId: 'h1abc123-STALE', content: '---\\ntitle: foo\\n---' })\` succeeds and pops a frontmatter overlay — even though the \`nodeId\` is bogus.

## Why this is inconsistent

The body-non-empty path correctly errors with \`NODE_NOT_FOUND\` on a stale \`nodeId\` and does NOT pop the overlay (per the round-3 fix in #508). The body-empty fall-through path doesn't have this guarantee.

## Why it's not critical

Agent intent is clearly the frontmatter when the call carries a frontmatter payload. The Option C path (\`nodeId: 'frontmatter'\`) is the documented happy path now — the body-empty fall-through is a B-era compat path. Probably benign.

## Proposed fix

In the frontmatter-only fall-through, require either:
- \`nodeId === 'frontmatter'\` (Option C), or
- the \`nodeId\` resolves to a real node in the doc (so the agent's intent is real, just frontmatter-only),

else return \`NODE_NOT_FOUND\` with guidance pointing at \`nodeId: 'frontmatter'\`.

## Refs

- PR #508 round 4 review — \"Frontmatter-only path bypasses nodeId validation\"
- The \`executeSuggestEdit\` frontmatter routing already covers Option B and Option C; this is consistency hardening for the Option B path.

mrangelmarino:
## Summary

- **#578 (P1 data-loss):** Added a defensive guard in `executeSuggestEdit` that blocks suggestions where the proposed replacement is less than 25% of the original node content (for nodes ≥200 chars). This surfaces a `SUGGESTION_DESTRUCTIVE` error instead of silently accepting a partial edit that would overwrite the majority of the node. The guard catches the exact failure mode: a typo-fix call on a single large paragraph emits only the first sentence (~4% of 1800 chars), and the guard stops it before the user can accept and lose everything.

- **#516 (consistency fix):** Tightened the frontmatter-only fall-through block (`if (frontmatterToStage && !content.trim())`) to validate the `nodeId` before staging the overlay. Previously, a bogus or stale `nodeId` would still pop the frontmatter overlay and return success. Now it returns `NODE_NOT_FOUND` with guidance pointing at `nodeId: 'frontmatter'`, matching the behaviour of the body-non-empty path.

## Root-cause analysis for #578

The data-loss had two layers:

1. **Single-node body structure**: the entire document body was stored as one large `<p>` node (~1800 chars) instead of discrete paragraph nodes. This happens when `acceptAISuggestion` accepts a replacement using `schema.text(suggestedText)` — always inline, never block-structured. Any content set this way, even with `\n` separating paragraphs, collapses into one paragraph node.

2. **Partial edit on a whole-node target**: with a single-body node, the agent's "fix the first-line typo" call targeted the only paragraph node. The model emitted just the corrected first sentence. Accepting it overwrote the whole node — gone.

**Structural root cause (single-paragraph body):** Confidently fixing this requires changing `acceptAISuggestion` to use `insertContentAt` (routed through the markdown parser) instead of `schema.text()`. That is a larger, higher-risk change separate from this P1 fix. The defensive guard prevents the data-loss consequence regardless of node structure.

## Test plan

- [ ] Open a multi-paragraph document. Call `suggest_edit` on a large paragraph node with a replacement that is only a single sentence. Confirm the tool returns `SUGGESTION_DESTRUCTIVE` (not a suggestion mark on the node).
- [ ] Call `suggest_edit` with a replacement of comparable length (e.g., a full rewrite of the same paragraph). Confirm the suggestion mark is applied normally.
- [ ] Call `suggest_edit` on a short node (<200 chars) with a very short replacement. Confirm the guard does not fire (short nodes are exempt).
- [ ] Call `suggest_edit` with `nodeId: 'frontmatter'` and frontmatter-only content. Confirm overlay pops as expected (happy path unaffected).
- [ ] Call `suggest_edit` with a stale/bogus `nodeId` and frontmatter-only content. Confirm `NODE_NOT_FOUND` is returned and no overlay pops.
- [ ] Call `suggest_edit` with a real `nodeId` (even if frontmatter-only content) — confirm that the overlay pops successfully (Option B compat path).

Closes #578
Closes #516

### DIFF
--- a/src/renderer/extensions/ai-suggestions/extension.ts
+++ b/src/renderer/extensions/ai-suggestions/extension.ts
@@ -6,10 +6,53 @@
  */
 
 import { Mark, mergeAttributes } from '@tiptap/core'
+import type { Editor } from '@tiptap/core'
+import type { Schema, Slice } from '@tiptap/pm/model'
+import { DOMParser as ProseMirrorDOMParser } from '@tiptap/pm/model'
 import type { MarkSerializerSpec } from 'prosemirror-markdown'
 import type { AISuggestionOptions, AISuggestionData, SuggestionType } from './types'
 import { useAnnotationStore } from '../ai-annotations'
 import { createWordDiffAnnotations } from '../../lib/diffUtils'
+
+/**
+ * Returns true when `text` contains block-level structure that would be lost
+ * if inserted as a flat schema.text() node — specifically, when it has at
+ * least one blank-line paragraph separator (the CommonMark block delimiter).
+ *
+ * Single-run content (a sentence, a phrase, inline edits) never contains
+ * `\n\n`, so the fast path is the overwhelmingly common case.
+ */
+function isMultiBlock(text: string): boolean {
+  return text.includes('\n\n')
+}
+
+/**
+ * Parse a markdown string into a ProseMirror Slice using the editor's
+ * configured tiptap-markdown parser, preserving block structure (paragraphs,
+ * headings, lists, etc.). Used only for multi-block acceptance so that
+ * inline accepts keep the existing schema.text() path byte-for-byte.
+ *
+ * Returns null if the editor doesn't have a markdown parser in storage
+ * (e.g., during tests) so callers can fall back to the text path.
+ */
+function parseMarkdownToSlice(editor: Editor, schema: Schema, text: string): Slice | null {
+  const parser = editor.storage?.markdown?.parser
+  if (!parser || typeof parser.parse !== 'function') return null
+
+  // Parse without inline:true so block structure (paragraph breaks, headings,
+  // etc.) is preserved. The parser renders to an HTML string.
+  const html = parser.parse(text) as string
+  if (typeof html !== 'string') return null
+
+  // Wrap in a <div> so ProseMirror's DOMParser sees a single root element.
+  const wrapper = document.createElement('div')
+  wrapper.innerHTML = html
+
+  // parseSlice preserves the block boundaries from the HTML and maps them to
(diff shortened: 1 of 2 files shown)


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/16b7e0ef-0d0c-4c4b-9575-cb09f40893a5?refs=b41edb5c0f3339482285ac09ccee351160340cb3&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
/review 

mrangelmarino:
/review 

Issue #531, opened 2026-05-18
HMR loses pending AI suggestions due to Editor remount


mrangelmarino:
## Symptom

When any renderer file changes during development (Vite HMR), pending AI suggestions disappear from the editor. The diff overlay is gone after the reload; the suggestion mark is no longer in the doc.

Production users never hit this (no HMR in built app), but it's painful during QA / iterative development on suggestion-adjacent code.

## Repro

1. `npm run dev`.
2. Open a doc.
3. Generate a pending suggestion: switch to Editor or Create mode, prompt the agent for an edit (or run \`/suggest_edit ...\`).
4. **Don't accept it.**
5. Touch any renderer file (e.g., \`echo >> src/renderer/index.css\`).
6. Vite reloads → pending suggestion is gone from the editor.

## Cause (suspected)

\`src/renderer/components/editor/Editor.tsx\` uses TipTap's \`useEditor\` with \`content: initialContent\`. \`initialContent\` is a one-shot \`useMemo\` computed at mount from \`document.content\`:

\`\`\`tsx
const initialContent = useMemo(() => {
  ...
  return getContentWithoutFrontmatter(document.content)
}, []) // Only run once on mount
\`\`\`

Suggestions are applied via \`editor.commands.setAISuggestion(...)\` after mount. The mark lands on the live editor's state but the marked-up content is never written back to the store's \`document.content\` until accept. On HMR remount, the editor is reconstructed from \`document.content\` — which doesn't have the post-mount marks — so pending suggestions are silently dropped.

## Possible fixes

Open question for whoever picks this up. Three angles:

1. **Persist suggestion marks back to the store before remount.** Add a teardown hook on the editor that serializes the current doc (with marks) into \`document.content\` so the next mount re-reads it. Risk: interacts with the autosave pipeline.

2. **Keep a parallel suggestion-data store and replay on mount.** \`useSuggestionStore\` already tracks suggestion data outside the editor (see \`src/renderer/extensions/ai-suggestions/store.ts\`). The mount path could iterate the store and re-apply each suggestion as a mark after the editor initializes. Risk: stale data if the underlying doc text changed.

3. **Make HMR not remount.** Probably not feasible — Fast Refresh's behavior on file changes is mostly out of our hands.

(1) is the most correct and probably the safest path.

## Out of scope

Production behavior is fine. This is dev-loop hygiene.

## Origin

[...]

PR #630, opened 2026-05-26
fix(editor): preserve pending AI suggestions across HMR remount (#531)

mrangelmarino:
## Summary

- On HMR, TipTap's editor is reconstructed from `document.content` in the store, which never has pending suggestion marks — those are applied post-mount and never written back. This caused suggestions to silently disappear on any renderer file touch during dev.
- Implements **option 2** from the issue: replay from the suggestion store on mount. On Editor cleanup (HMR unmount), a new `snapshotSuggestions()` action captures the live marks from ProseMirror into `useSuggestionStore.pendingSuggestions`. The existing `restoreAISuggestions` `useEffect` (already present for the tab-switch case) picks them up on the next mount and calls `restoreAISuggestions()`.
- No IndexedDB writes, no autosave pipeline interaction — it's a pure in-memory snapshot that lives only for the duration of the HMR cycle.

## Why option 2 over option 1

Option 1 (persist marks back to `document.content` before remount) would serialize the marked-up document into the store's content field, entangling the autosave pipeline and potentially persisting internal mark state to disk. Option 2 reuses the existing store + restore machinery already used for tab switches — the blast radius is two small, well-scoped changes.

## Guards against double-apply and stale data

- **Double-apply**: `clearSuggestions()` is called inside the existing restore `useEffect` immediately after replay, so the snapshot is consumed once.
- **Stale data**: `restoreAISuggestions` searches for `suggestion.originalText` in the document before applying. If the text changed (e.g., the user edited between HMR cycles), the search fails and the suggestion is skipped with a `console.warn` — no crash, no silent corruption.

## How to verify (dev-only repro)

1. `npm run dev`
2. Open a document with some text
3. Use the chat agent to request an edit (or run `suggest_edit` via MCP) — a pending suggestion highlight appears
4. Do NOT accept it
5. Touch any renderer file: `echo >> src/renderer/index.css`
6. Vite HMR fires → the suggestion highlight **survives** (was lost before this fix)
7. Accept/reject works normally after restore

## Files changed

- `src/renderer/extensions/ai-suggestions/store.ts` — adds `snapshotSuggestions()` action (in-memory only, no IndexedDB)
- `src/renderer/components/editor/Editor.tsx` — adds snapshot call in editor-instance registration effect cleanup

Closes #531

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/renderer/components/editor/Editor.tsx
+++ b/src/renderer/components/editor/Editor.tsx
@@ -373,11 +373,25 @@ export function Editor() {
     }
   }, [editor, document.content, document.documentId])
 
-  // Register editor instance in store for cross-component access
+  // Register editor instance in store for cross-component access.
+  // On cleanup (HMR remount or unmount), snapshot any live AI suggestions into
+  // the suggestion store so the next mount can replay them via the existing
+  // restoreAISuggestions useEffect — without touching the autosave pipeline.
   useEffect(() => {
     setEditorInstance(editor)
-    return () => setEditorInstance(null)
-  }, [editor, setEditorInstance])
+    return () => {
+      setEditorInstance(null)
+      // Snapshot live suggestions for HMR replay (option 2 from issue #531).
+      // Only runs when there is an editor with an active document — skips
+      // clean unmounts where document.documentId is empty (e.g., empty state).
+      if (editor && document.documentId) {
+        const liveSuggestions = getAISuggestions(editor)
+        if (liveSuggestions.length > 0) {
+          useSuggestionStore.getState().snapshotSuggestions(document.documentId, liveSuggestions)
+        }
+      }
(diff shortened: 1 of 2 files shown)


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/9b358f8e-797b-4ed6-a63e-d902f4d91989?refs=8cbd01b4c4ddd2022e34dc1a125823e64aa9f2b2&source=prChecksComment) |

mrangelmarino:
/review 

Issue #646, opened 2026-05-28
Implement "Reopen Closed Tab" on Cmd+Shift+T


mrangelmarino:
## Summary

Implement the standard **"Reopen Closed Tab"** behavior on **Cmd+Shift+T**.

During the 1.5.0 cut, `Cmd+Shift+T` was freed up — it previously triggered "Convert to Plain Text," which was an unintuitive hijack of the near-universal "reopen closed tab" shortcut. Convert-to-Plain-Text moved to `Cmd+Option+T` (and Convert-to-Markdown to `Cmd+Option+M`). `Cmd+Shift+T` is now **unbound**, reserved for this feature.

## Current state

There is **no closed-tab history**. `tabStore` only has `removeTab` (`src/renderer/stores/tabStore.ts`), and `useTabs` (`closeTab`/`forceCloseTab`) calls it without retaining anything. So this needs to be built, not just rebound.

## Proposed implementation

- **`tabStore`**: add a bounded `closedTabs` stack (e.g. cap 10). In `removeTab`, snapshot the closing tab's full restorable state before deleting (path, in-memory content, `documentId`, frontmatter, emoji, dirty flag — enough to restore unsaved edits, not just reload from disk).
- Add a `reopenLastClosedTab()` action that pops the stack, recreates the tab, and activates it.
- **Wiring**: add a "Reopen Closed Tab" File-menu item with accelerator `CmdOrCtrl+Shift+T` → `sendMenuAction('reopenTab')` (`menu.ts`) → `App.tsx` `onMenuAction` case → `useTabs.reopenLastClosedTab()`.

## Design questions to settle

- **Unsaved content**: restore the in-memory content (preferred) vs. reload from `path`? A tab closed with unsaved edits should come back with those edits.
- **documentId**: reuse the original id (to rejoin chat history / persistence) vs. regenerate?
- **Session persistence**: should the closed-tab stack survive app restart (IndexedDB), or be session-only?
- **Preview tabs**: skip single-click preview tabs, or include them?
- **Edge cases**: closing the last tab, reopening a tab whose file was deleted on disk.

## Context
Deferred from the 1.5.0 menu/shortcut cleanup (kept the release scoped to a bug batch + HTML export). Not urgent.

Issue #645, opened 2026-05-28
Investigate AI annotation fade-out: annotations become effectively invisible over time


mrangelmarino:
## Summary

AI annotations (the time-fading provenance highlights on AI-generated text) fade to **near-invisibility** over their lifetime. By design they should *gently* fade but **remain at least slightly visible** so the AI provenance stays discoverable — the only thing that should make them disappear is the user explicitly toggling annotation visibility off.

## Current behavior

Fade is driven by `calculateOpacity()` in `src/renderer/types/annotations.ts`:

```ts
export const ANNOTATION_CONSTANTS = {
  MAX_AGE_MS: 7 * 24 * 60 * 60 * 1000, // fully faded after 7 days
  MIN_OPACITY: 0.05,                   // floor
  REFRESH_INTERVAL_MS: 5 * 60 * 1000,
}

export function calculateOpacity(createdAt, maxAge = MAX_AGE_MS) {
  const age = Date.now() - createdAt
  if (age <= 0) return 1
  if (age >= maxAge) return ANNOTATION_CONSTANTS.MIN_OPACITY
  const ratio = age / maxAge
  const opacity = Math.pow(1 - ratio, 0.7) // exponential decay
  return Math.max(ANNOTATION_CONSTANTS.MIN_OPACITY, opacity)
}
```

- The opacity floor is **`MIN_OPACITY = 0.05`** (5%), which is effectively invisible. As an annotation approaches `MAX_AGE_MS` (7 days) it decays toward 0.05 and reads as "gone."
- `src/renderer/extensions/ai-annotations/plugin.ts` (two sites, ~L131 and ~L241) additionally **skips rendering** the decoration entirely when `opacity < MIN_OPACITY` — so anything at/below the floor is dropped.

Net effect: a week-old AI edit's annotation is visually indistinguishable from un-annotated text, even though the user never toggled annotations off.

## Desired behavior

- Annotations should **retain a perceptible minimum opacity** over time (e.g. raise the floor to ~0.2–0.35, and/or flatten the decay curve) so AI provenance stays visible at a glance.
- The **only** way an annotation should disappear is the user's explicit visibility toggle (`annotationsVisible` / `toggleVisibility` → the `hide-annotations` class on the editor), not the passage of time.

## Investigation / open questions

- What's the right minimum opacity floor for "still clearly visible but visually de-emphasized"?
- Should there be a full-fade-to-floor over 7 days at all, or a gentler decay that settles at a persistent floor much sooner?
[...]

PR #641, opened 2026-05-27
feat(export): HTML export with embedded markdown for round-trip editing

robbylucia:
## Summary

- **Export HTML**: New "Export HTML..." option in the toolbar menu. Produces a standalone HTML file with clean styling (dark mode support, task lists, tables, code blocks).
- **Round-trip editing**: The original markdown is base64-encoded in a `<script type="application/x-prose-markdown">` tag inside the HTML. Opening a Prose HTML file in Prose extracts and loads the markdown for continued editing.
- **Image inlining**: Local images are converted to base64 data URIs in both the visible HTML and the embedded markdown, so exported files are fully portable.
- **Open dialog**: File filter updated to show markdown and HTML files together by default ("Documents" filter).

### New IPC handlers
- `file:exportHtml` — save dialog filtered to `.html`
- `file:readBase64` — read binary files as base64 (for image inlining)

### Files changed
- `src/renderer/lib/htmlExport.ts` (new) — export builder, image inlining, markdown extraction
- `src/main/ipc.ts` — 2 new handlers, updated open dialog filter
- `src/preload/index.ts` — expose new IPC methods
- `src/renderer/types/index.ts` — type definitions
- `src/renderer/lib/browserApi.ts` — web mode fallback
- `src/renderer/hooks/useEditor.ts` — detect `.html` on open, extract embedded markdown
- `src/renderer/components/layout/Toolbar.tsx` — Export HTML menu item

## Test plan

- [ ] Open a document with text, images, task lists, tables, and code blocks
- [ ] Click ⋯ menu → Export HTML → save the file
- [ ] Open the HTML in a browser — content should render with styling, images should display
- [ ] Open the same HTML file back in Prose (Open... → select the .html file) — original markdown should load with images intact
- [ ] Verify vanilla HTML files (without the Prose tag) are rejected gracefully on open
- [ ] Test with an unsaved document (no images on disk) — export should still work, images fall back to data URIs if already base64

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #635, opened 2026-05-27
Harden all AI entry points when no LLM is configured (visible feedback + no data loss + disabled controls)


mrangelmarino:
## Background

This is the **original audit** requested on #631 ("make sure *nothing* breaks if a user hasn't configured an LLM"). #632/#633 fixed the narrow symptom — the **Process** button silently deleting a comment — but not the root problem.

Two things surfaced during review of #631:
- @robbylucia: *"if it displayed an error I didn't notice it. Errors might've errored!"* — the only user-facing "AI not configured" error lands **in the chat panel**, which may be collapsed/hidden, so the user sees nothing.
- The codebase uses **three different config checks** with three different behaviors.

## The inconsistency

| Entry point | Current guard | Behavior |
|---|---|---|
| `sendMessage` (chat, comment Process, Process All, suggestion replies) | `validateConfig(settings.llm)` | Error message **in chat panel only** |
| `generateSummary` (auto-fired on doc open, 7+ call sites) | `!settings.llm.apiKey` | Silent no-op |
| `ChatPanel` | `hasApiKey` | — |

No single source of truth for "is AI usable," and the only user-initiated error path is invisible when the chat panel isn't open.

## Goals

1. **Single source of truth** — one `isAIConfigured()` / `useAIConfigured()` helper wrapping `validateConfig`, used everywhere.
2. **No data loss** — comments, suggestions, and any other user content are never destroyed when AI was never dispatched (baseline landed in #633; extend/standardize).
3. **Visible feedback** — user-initiated AI actions show a **toast/notification** ("AI not configured — add an API key in Settings ⌘,") regardless of chat-panel visibility.
4. **Disabled controls** — proactively disable/grey-out AI controls (comment **Process**, **Process All**, suggestion-reply processing, chat **Send**) with an explanatory tooltip when AI isn't configured.
5. **Auto-fired calls** (summaries, etc.) silently no-op via the same shared check.
6. Confirm the reMarkable OCR path (needs an Anthropic key) degrades gracefully.

## Out of scope

- Web-mode CORS limitations (AI unavailable in browser is a separate known constraint).

## Acceptance

- With no API key set: every AI control is either disabled-with-tooltip or shows a visible toast; no comment/suggestion is ever lost; no failing LLM calls fire automatically.
- Ships in **v1.4.1**.

Refs #631

PR #647, opened 2026-05-28
Release v1.5.0: Wave 0 bug/QoL batch + HTML export

### LINKED ISSUE #516: fix(editor): tighten suggest_edit frontmatter-only fall-through to require nodeId='frontmatter'

### LINKED ISSUE #517: feat(files): handle chokidar 'change' events to refresh modifiedAt on external saves
## Context

The File Explorer watcher (\`src/main/fileWatcher.ts\`) currently subscribes to \`add\`, \`addDir\`, \`unlink\`, and \`unlinkDir\` events only. External saves that mutate file content without renaming (the typical case for \`echo >> file\`, IDE saves outside Prose, etc.) fire chokidar's \`change\` event — which we don't handle. So a file's listed \`modifiedAt\` won't refresh in the explorer until the user navigates away and back.

Flagged by the #509 round 4 review as \"edge case; probably fine to leave.\" Not a regression; surfaces now only because we have a watcher at all.

## Proposed

Wire up \`watcher.on('change', ...)\` to emit a \`{ type: 'changed', path }\` event — or reuse \`'created'\` since the renderer just reloads on any event, and add the new type to the union if we want renderers to distinguish later.

Renderer side: same debounced \`loadFiles()\` reload path. Mostly a one-liner in fileWatcher.ts + a type union update.

## Refs

- PR #509 round 4 review (\"Minor: \`change\` events not handled\")

### LINKED ISSUE #518: feat(files): watcher misses events inside expanded subfolders (depth-0 limitation)
## Context

The File Explorer watcher (\`src/main/fileWatcher.ts\`) uses \`chokidar.watch(path, { depth: 0 })\`, which watches the root and its **direct** children only. Subfolders aren't watched recursively.

That's intentional for cost reasons — recursive watching of large trees fires excessive events. But it creates a UX surprise: if a user has \`Notes/2026/\` expanded in the file tree and runs \`touch Notes/2026/newfile.md\` from the terminal, the watcher won't see it. The new file won't appear in the explorer until the user collapses and re-expands \`Notes/2026/\`.

Flagged by the #509 round 5 review.

## Options

1. **Expand watcher depth dynamically.** Track \`expandedFolders\` in the main process; for each expanded path, attach an additional shallow watcher. Tear down as folders collapse. Bounded by user's actual expansion state, not the whole tree.

2. **Restart watcher with \`depth: N\`** based on the max expansion depth. Simpler but wastes work for narrow trees.

3. **Document the limitation** in the File Explorer (e.g., tooltip or hover hint on the refresh affordance) and leave the behavior. Lowest cost.

Option 1 is the most correct and bounded. Option 3 is the cheapest. Open to discussion.

## Refs

- PR #509 round 5 review
- #517 — related: \`change\` events for mtime refresh (also watcher gap)

### LINKED ISSUE #531: HMR loses pending AI suggestions due to Editor remount

### LINKED ISSUE #539: Source mode selection highlight too bright in dark/monochromatic theme
## Summary

In source mode, selecting text in the default dark (\"monochromatic\") theme paints the selection in a washed-out pale lavender that's harder to look at than the surrounding text. The selection is brighter than the body text, which inverts the usual visual hierarchy and makes the highlighted region the most attention-grabbing thing on screen.

## Repro

1. Switch to the default dark theme
2. Open any document in **source** mode (the markdown view, not the prose view)
3. Select a paragraph or two

The selected region shows up as a near-white wash that fills the line height fully — see screenshot attached to the originating conversation.

## Root cause

`src/renderer/components/editor/SourceEditor.tsx:96-98`:

```ts
'&.cm-focused .cm-selectionBackground, .cm-selectionBackground': {
  backgroundColor: 'hsl(var(--primary) / 0.15)',
},
```

In dark mode, `--primary` is `0 0% 98%` (near-white — `src/renderer/index.css:36`). White at 15% alpha over the dark blue-tinted background (`hsl(240 10% 3.9%)`) renders as a luminous lavender that competes with the foreground text. The same issue lurks in the prose editor's `::selection` rule (`src/renderer/index.css:414-416`) using `hsl(var(--primary) / 0.2)` — even brighter there, just less visible because TipTap selections are usually shorter than source-mode ones.

In the light theme, `--primary` is dark (`240 5.9% 10%`), so the same alpha formula produces a soft gray that works fine. The bug is specifically that the formula doesn't degrade gracefully when `--primary` is near-white.

## Proposed fix

Introduce a dedicated `--selection` token per theme rather than deriving from `--primary`. This is the right level of abstraction — selection color is a UI concern that doesn't always want to track the brand/primary color.

```css
/* index.css */
:root { --selection: 240 5.9% 10% / 0.18; }       /* light: dark primary works */
.dark { --selection: 0 0% 70% / 0.22; }            /* dark: muted gray, not pure white */
.termy-green-dark { --selection: 120 60% 40% / 0.3; }
.termy-green-light { --selection: 120 60% 25% / 0.18; }
```

Then both call sites become:
```css
::selection { background: hsl(var(--selection)); }
```
```ts
'&.cm-focused .cm-selectionBackground, .cm-selectionBackground': {
  backgroundColor: 'hsl(var(--selection))',
},
```

## Acceptance

- [ ] Selected text in source mode is legible (selection is dimmer than foreground text) across all four themes: `dark`, `light`, `termy-green-dark`, `termy-green-light`
- [ ] Prose-mode selection (`::selection` rule) uses the same token and looks visually consistent with source mode
- [ ] Selection is still clearly distinguishable from non-selected text — fix the contrast in both directions, don't just dim until invisible

### LINKED ISSUE #561: UI: hide AI model selector when AI is disabled
## Description

When a user chooses "Continue without AI" during onboarding, the AI model selector is still visible in the bottom-right status bar. While it may not be functionally active, its presence contradicts the user's intent to avoid AI features entirely.

## Expected Behavior

If the user has disabled AI (or skipped AI setup), the model selector in the status bar should be hidden to respect the "no AI" choice.

## Current Behavior

The model selector chip remains visible in the bottom-right corner regardless of AI configuration state.

## Source

External user feedback (coworker demo session, 2026-05-22).

## Labels

- `bug`
- `ux`

### LINKED ISSUE #562: fix(editor): comment tooltip clipped by top nav bar
## Description

When hovering over a comment that is positioned near the top of the document, the tooltip/popover is partially hidden behind the top navigation bar.

## Expected Behavior

Comment tooltips should reposition (e.g., render below the comment instead of above) when there isn't enough space above, or the tooltip should be rendered in a portal layer that isn't clipped by the nav.

## Current Behavior

The comment hover tooltip renders above the comment and gets cut off by the top navigation bar, making it unreadable.

## Source

External user feedback (coworker demo session, 2026-05-22).

### LINKED ISSUE #566: feat(editor): syntax highlighting in code blocks
## Description

Add syntax highlighting for fenced code blocks in the editor. Currently code blocks render as plain monospace text. Language-aware highlighting would improve readability and make Prose more competitive with dedicated markdown editors.

## Considerations

- TipTap has community extensions for code highlighting (e.g., `@tiptap/extension-code-block-lowlight` with lowlight/highlight.js)
- Bundle size impact should be evaluated — highlight.js with all languages is heavy. Consider loading only common languages or lazy-loading grammars
- Should respect the current theme (dark mode)

## Source

External user feedback (coworker demo session, 2026-05-22).

### LINKED ISSUE #571: insert(position=after_node) on a heading absorbs text into the heading node instead of creating a sibling paragraph
## Summary

In Create Mode, `insert` with `position=after_node` anchored on a **heading** nodeId places the new text **inside the heading node** (it takes on heading styling) instead of creating a sibling paragraph after the heading.

## Repro (from #467 Stage 6 QA)

1. Create Mode, a document with headings (e.g. `# Test Document`).
2. Ask the agent: "Add a short closing paragraph summarizing this."
3. Agent calls `insert` with `position=after_node` on a heading nodeId.
4. **Observed:** the closing paragraph is absorbed into a heading node (renders as a heading). The agent itself then notices "the closing paragraph was inserted as part of the heading node" and has to correct it.
5. **Expected:** a new paragraph node is created immediately after the heading, as a sibling block.

## Leads

- `src/renderer/lib/tools/executors/editor.ts` — `executeInsert`, the `after_node` / `before_node` handling (added in #546).
- Insert position may resolve to the end of the heading's inline content rather than the document level between block nodes, so parsed markdown merges into the heading.

## Notes

Pre-existing; surfaced during #548 testing but unrelated to the annotation change. Insert-anchor positions landed in #546. Base: `release/v1.2-agent-persona`.

Refs #467 #546

### LINKED ISSUE #572: Full-node edit clears AI annotations on unchanged adjacent words in the same node
## Summary

A full-node `edit` clears existing AI annotations on **unchanged adjacent words** in the same node. Editing one word drops the provenance highlight on a neighbor that wasn't touched.

## Repro (from #467 Stage 6 QA)

1. Create Mode. A sentence begins "A good ..." where "A" already carries an AI annotation from a prior edit.
2. Ask the agent to change "good" → "excellent".
3. Agent `edit`s the node → "A excellent ...".
4. **Observed:** the annotation on "A" is lost; only "excellent" ends up annotated.
5. **Expected:** the unchanged "A" keeps its prior annotation; the word-diff re-applies annotations only to changed spans without clobbering untouched neighbors.

## Leads

- `edit` replaces the entire node content (`executeEdit`), so decorations over the old range may be invalidated instead of position-mapped.
- `src/renderer/extensions/ai-annotations/store.ts` — position remap / split logic (`mapPos`, insertion-split handling).
- `src/renderer/lib/diffUtils.ts` — `createWordDiffAnnotations` re-annotates the changed range; interaction with pre-existing annotations in the same node.

## Notes

Pre-existing; `edit` used this annotation path before #548. Surfaced during #548 testing. Base: `release/v1.2-agent-persona`.

Refs #467

### LINKED ISSUE #595: Comments are dropped on tab switch (and any editor state reset) — no persistence path
## Summary
Comments added to a document silently disappear when the user switches tabs (or triggers any change that resets the editor state). The comment marks exist only in the in-memory TipTap editor state and are never persisted, so the moment the editor is repopulated from saved content they're gone — with no warning and no way to recover them.

## Reproduction
1. Open a document, select some text, add a comment.
2. Switch to another tab (or back, or trigger an editor content reset).
3. Return to the document → the comment is gone.

## Where
- `src/renderer/extensions/comments/extension.ts:16-21` — `commentMarkdownSerializer` emits empty `open`/`close` strings, so comment marks are **stripped entirely** when content is serialized to markdown.
- `src/renderer/components/editor/Editor.tsx:228-246` — the debounced `onUpdate` handler stores `editor.storage.markdown.getMarkdown()` into `editorStore`. Because of the serializer above, that string already has no comment marks.
- `src/renderer/hooks/useTabs.ts` (`saveCurrentTabState` ~L104-109) — saves `document.content` (the comment-free markdown) back onto the tab. There is no `saveComments` step. Contrast with AI suggestions, which *do* have `saveSuggestions`/`loadSuggestions`.
- `src/renderer/components/editor/Editor.tsx:306-321` — on tab switch the editor does a hard `EditorState.create(...)` + `view.updateState(...)` reset, rebuilding the doc from the (comment-free) markdown. The AI-suggestion restore at L427-446 (`restoreAISuggestions`) has **no comment equivalent**.

## Root cause
Comments are implemented as a TipTap `comment` mark (`data-comment-id` / `data-comment` / `data-comment-created`) that lives **only** in ProseMirror in-memory state. There is:
- no `comments` object store in `src/renderer/lib/persistence.ts`,
- no `comments` field on the `Tab` / `TabDraft` interfaces,
- no save/restore path on tab switch.

The serializer deliberately emits nothing, so comments don't round-trip through markdown either. Persistence was apparently never implemented for this mark (unlike AI suggestions, which have a full IndexedDB save/restore path).

## Fix direction
Mirror the AI-suggestions persistence pattern (the utilities mostly already exist):
1. Add a `comments` IndexedDB store in `persistence.ts` with `saveComments`/`loadComments`.
2. In `useTabs.ts:saveCurrentTabState()`, capture marks via the existing `getComments(editor)` util (`extension.ts:186-234`) and persist them keyed by tab/document.
3. Add a `restoreComments` editor command and call it after tab switch in `Editor.tsx`, alongside the existing `restoreAISuggestions` block (L427-446).

Alternatively, if comments should travel with the file, change `commentMarkdownSerializer` to emit inline HTML (`<span data-comment-id=...>`) so they round-trip through the saved markdown — but that changes on-disk file content, so the per-tab/IndexedDB approach is the safer match for current behavior.

## Severity
**High** — silent, unrecoverable loss of user-authored editorial content during an extremely common interaction (switching tabs).

### LINKED ISSUE #642: fix(editor): AI annotation toggle shows active for unannotated text
Found during the Wave 0 HITL-QA sweep on `release/wave-0`.

## Symptom
Selecting text that has **no** AI annotation shows the AI-annotation toggle (the Bot button in the selection popover) in the active / "annotated" state. Clicking it operates across the whole selection, and the provenance/reason surfaced belongs to a *nearby* annotation (e.g. the last AI edit in an adjacent paragraph). Toggling off trims that neighbour's annotation across the selected span.

## Root cause
`src/renderer/components/editor/SelectionPopover.tsx` computed `hasAIAnnotation` with a loose any-overlap test:

```ts
getAnnotationsInRange(annotations, from, to).length > 0   // a.from < to && a.to > from
```

AI annotations are inline decorations driven by the store's `from`/`to`. Any annotation whose range overlaps the selection by even one character — including one whose mapped range reaches into the selected span from an edit elsewhere — lit the toggle up. The remove path (`removeAnnotationsInRange`) used the same loose predicate, so toggling off trimmed the overlapping neighbour across the span.

## Status
**Pre-existing** — `SelectionPopover.tsx` and `ai-annotations/store.ts` are byte-identical to `main`, so this reproduces on `main` too. Wave 0's #572 (annotation preservation on edits) can *amplify* it by leaving more annotations whose mapped ranges overlap adjacent text, which is how it surfaced during QA.

## Fix
Make the toggle reflect "is the **selected text** annotated?" as a *coverage* question — the selection must be fully covered by annotation(s) — rather than merely touched. Scoped to `SelectionPopover.tsx`; the shared `getAnnotationsInRange` is left untouched (it backs legitimate overlap/decoration lookups elsewhere). Verified via live HMR on `release/wave-0`: the toggle now reads inactive on unannotated text and active on annotated text.

mrangelmarino:
## Prose v1.5.0

The reconciled **Wave 0 bug/QoL batch** plus **HTML export** (contributed by @robbylucia), with a few menu/shortcut cleanups. Bumps `1.4.1 → 1.5.0`.

### Highlights
- **HTML export with round-trip editing** (#641, @robbylucia) — export a standalone, styled HTML file with the source markdown embedded so it re-opens in Prose for continued editing; local images inlined as data URIs. Available from the toolbar ⋯ menu **and** the File menu.
- **Wave 0 bug/QoL batch** — 13 fixes across the editor, comments, AI annotations, file watching, and the status bar (see Closes list).
- **Menu/shortcut cleanup** — "Export as HTML…" added to the File menu; freed `Cmd+Shift+T` (it was an unintuitive convert-to-txt hijack of the standard *reopen closed tab* shortcut) by moving **Convert to Plain Text → `Cmd+Option+T`** and **Convert to Markdown → `Cmd+Option+M`**; dropped the colliding insert-table shortcut (tables still come from markdown pipe syntax).
- **Security** — pin `tmp ≥ 0.2.6` via `overrides` to clear Dependabot #74 (dev-only transitive path-traversal advisory, GHSA-ph9p-34f9-6g65).

### Closes
Closes #578
Closes #516
Closes #595
Closes #531
Closes #572
Closes #571
Closes #562
Closes #539
Closes #517
Closes #518
Closes #566
Closes #561
Closes #642

> #563, #564, and #631 were fixed independently on `main` during this batch and are already closed — intentionally **not** re-closed here.

### Deferred follow-ups (tracked, not in this release)
- #645 — AI annotation fade-out becomes effectively invisible over time
- #646 — implement "Reopen Closed Tab" on the now-free `Cmd+Shift+T`
- Non-Prose HTML files open as a silent no-op; should surface a notification (`notificationStore`)
- Modernize the HTML-embed base64 codec with a chunked `TextEncoder`/`TextDecoder`

### Verification
- `npm run build` clean; `tsc --noEmit` shows only pre-existing project-reference errors (TS6305/6306/6310).
- HITL QA passed on the assembled branch: HTML export round-trip, status-bar count/cursor reconciliation (#563/#564 kept main's implementation), #561 model-selector-hide + #635/#636 AI-off hardening coexistence, and a Wave 0 batch spot-check.

### Reviewer note
This is an umbrella release branch. Child-PR work was reconciled against `main` (incl. the concurrent #633/#635/#636 fixes) — review against the `release/wave-0` base, and discount any flags about foundation types/code already present on `main`.


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/cf79b244-8761-475f-a4aa-617854a187cd?refs=40331db8833b6bf4e679c200ad324b59284b88c9&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
/review 

mrangelmarino:
Merging for the **v1.5.0** cut (maintainer-authorized autopilot).

- CI green; **Electron security checklist all-pass**; the prior review's one functional bug (`handleExportHtml` not memoized) is fixed in `40331db` via `useCallback`.
- Remaining review items are **non-blocking**: `restoreComments` first-occurrence + block-boundary separator edge cases are inherent limitations of the (net-improvement) #595 text-search restoration in already-QA'd code → tracked in #648; `file:readBase64` reading any validatePath'd path is explicitly *not a regression* (same as `file:read`); non-Prose-HTML silent open is a tracked deferral.
- HITL QA passed earlier on the assembled branch.

Cutting now → tag `v1.5.0`.

PR #745, opened 2026-06-13
fix(editor): preserve caret on click when promoting preview tab (#729)

### LINKED ISSUE #729: Cursor & scroll jump to top on first click into a document body (preview-tab open)
## Summary

Clicking into a document body **for the first time** after opening and scrolling jumps the caret — and the viewport — to the **top** of the document, instead of landing where you clicked.

## Repro

1. Single-click a file in the explorer to open it (opens as a **preview** tab).
2. Scroll down and read **without clicking into the editor body**.
3. Click into the body somewhere mid-document.
4. ❌ The caret and scroll position jump to the very top of the document.

Subsequent clicks behave correctly — only the *first* click after open misbehaves.

## Root cause

`src/renderer/components/editor/Editor.tsx:532-552` — a one-shot "focus on document load" effect:

```ts
useEffect(() => {
  if (hasFocusedRef.current) return
  if (useEditorStore.getState().isPreviewTab) return   // early-return BEFORE setting the ref
  const shouldShowEmptyState = !isEditing && !document.path && !document.content && !document.isDirty
  if (editor && !shouldShowEmptyState) {
    hasFocusedRef.current = true
    const timer = setTimeout(() => {
      editor.commands.focus('start')                   // ← forces caret to doc position 1
      const editorContainer = editor.view.dom.closest('.overflow-auto')
      if (editorContainer) editorContainer.scrollTop = 0   // ← forces viewport to top
    }, 50)
    return () => clearTimeout(timer)
  }
}, [editor, isEditing, document.path, document.content, document.isDirty])
```

The `'start'` argument and the `scrollTop = 0` are intended to open a fresh document at the top with the caret ready. The bug is **when** that fires for preview tabs:

- On a **preview** open, `isPreviewTab` is `true`, so the effect **returns early — without setting `hasFocusedRef.current = true`**. The one-time focus is therefore *deferred*, and the guard stays un-tripped while you read/scroll.
- Your first click into the body **promotes** the preview tab to a permanent tab (`isPreviewTab → false`) and changes a dependency of this effect, so it **re-runs** — now both guards pass, `hasFocusedRef` is still `false`, and it fires `focus('start')` + `scrollTop = 0`. The deferred "open at top" initialization lands *on top of your click*, throwing caret and scroll to the top.
- After that run, `hasFocusedRef.current` is `true`, so every later click is a no-op — matching "only the first click."

**Prediction to confirm:** the bug should reproduce **only for preview-opened (single-click) documents**, not for double-click / permanent opens — because a permanent open runs the focus-to-top harmlessly on open (when `isPreviewTab` is already `false`), leaving later clicks untouched.

## Secondary suspect

`Editor.tsx:343-377` — the content-sync effect (`setContent` / `updateState` + `scrollTop = 0`) also resets selection to start and scroll to top. If a content update lands after the first click it could produce the same symptom. Lower likelihood for this repro (no edit has happened yet on first click), but worth ruling out.

Note: there is **no** "restore saved cursor position" path involved — `editorStore.setCursorPosition` only drives the status-bar display; `Tab.cursorPosition` is never re-applied to the ProseMirror selection. So this is purely the forced `focus('start')`, not a bad restore.

## Proposed fix (decide during implementation)

The goal: a fresh open lands at top with caret ready, but a user's *click-to-promote* must not be overridden.

- **Trip the guard on the preview early-return** — set `hasFocusedRef.current = true` (or otherwise mark "initial focus handled") even when `isPreviewTab` short-circuits, so promotion doesn't re-fire `focus('start')`. The click then keeps its natural caret/scroll.
- and/or **use `focus()` instead of `focus('start')`** (and drop the forced `scrollTop = 0`) when the focus is the result of a user click/promotion — preserve the existing selection rather than resetting it.

Keep the "open fresh doc at top" behavior for genuinely new/just-opened docs; only avoid clobbering an interaction-driven focus.

## Acceptance criteria
- [ ] Open a doc (preview, single-click), scroll, click mid-document → caret lands at the click; no jump to top
- [ ] Fresh opens still start at the top with the caret ready
- [ ] Double-click/permanent opens unchanged
- [ ] Verified live via Circuit Electron for both preview and permanent open paths

## Notes
- Surfaced by user QA on 2026-06-11. Companion to the Do-Next editor/UX work; this is a focus-timing bug, not a data issue.

mrangelmarino:
Closes #729

## What & why
Clicking into a document body for the first time after opening it as a **preview tab** caused the caret and viewport to jump to the top, discarding the click position.

**Root cause:** the auto-focus effect in `src/renderer/components/editor/Editor.tsx` returned early when `isPreviewTab` was true *without* marking `hasFocusedRef.current = true`. When the user clicked mid-document, `promoteCurrentPreview()` flipped the tab to permanent (`isPreviewTab → false`), the effect re-ran with `hasFocusedRef` still `false`, and it executed `editor.commands.focus('start')` + `scrollTop = 0` — overriding the caret/scroll the click had just set.

**Fix:** mark `hasFocusedRef.current = true` in the preview short-circuit so a later promotion no longer re-triggers the one-shot focus/scroll. Fresh permanent opens (double-click) are unaffected because `isPreviewTab` is `false` from the start, so the focus-at-top behavior still runs once.

Single-file, renderer-only change. No privilege-boundary or hot-file touches.

## Manual QA
1. Open a document from the explorer with a single click (preview tab) — content long enough to scroll.
2. Without clicking into the body, scroll down to the middle of the document.
3. Click into the body mid-document.
   - **Expected:** caret lands where you clicked; no jump to the top.
4. Open a fresh document via double-click (permanent tab).
   - **Expected:** it starts at the top with the caret ready (unchanged behavior).
5. Re-open the same preview document and confirm double-click/permanent opens behave as before.

_Conversation: https://app.warp.dev/conversation/e2b7168a-5330-4196-a697-605dc105526f_
_Run: https://oz.warp.dev/runs/019ec2f7-f76b-7c23-9808-c131e490844b_

_This PR was generated with [Oz](https://warp.dev/oz)._

### DIFF
--- a/src/renderer/components/editor/Editor.tsx
+++ b/src/renderer/components/editor/Editor.tsx
@@ -532,8 +532,13 @@ export function Editor() {
   useEffect(() => {
     // Only focus once per document load
     if (hasFocusedRef.current) return
-    // Don't steal focus during preview tab navigation
-    if (useEditorStore.getState().isPreviewTab) return
+    // Don't steal focus during preview tab navigation. Mark focus as handled so a
+    // later promotion to a permanent tab (e.g. the user clicks mid-document) doesn't
+    // re-run this effect and yank the caret/scroll back to the top (#729).
+    if (useEditorStore.getState().isPreviewTab) {
+      hasFocusedRef.current = true
+      return
+    }
 
     const shouldShowEmptyState = !isEditing && !document.path && !document.content && !document.isDirty
     if (editor && !shouldShowEmptyState) {


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/a149a35a-15e6-4320-be5a-04b26ecc8466?refs=50d9cd6d6e1df801a81f83762c84488fdb00e8e6&source=prChecksComment) |

mrangelmarino:
/review 

Issue #391, opened 2026-03-30
fix(sentry): CSP blocks sentry-ipc protocol in MAS build


mrangelmarino:
## Problem

The MAS (TestFlight) build shows heavy console spam from Sentry CSP violations:

```
Fetch API cannot load sentry-ipc://scope/sentry_key.
Refused to connect because it violates the document's Content Security Policy.
```

Additionally:
- `sentry:setEnabled` IPC throws: "Sentry SDK should be initialized before the Electron app 'ready' event is fired"
- The Settings prompt to enable error reporting may not be visible

## Root cause

The renderer's CSP `connect-src` directive doesn't include the `sentry-ipc:` custom protocol that `@sentry/electron` uses for renderer↔main IPC. The SDK initializes in the renderer but can't communicate with the main process, causing repeated failed fetch attempts.

The initialization timing error suggests `initSentry()` in main may not be running early enough relative to the renderer's `sentry:setEnabled` IPC call.

## Fix candidates

1. **Add `sentry-ipc:` to CSP** — Update the `connect-src` directive in the renderer HTML or main process `session.webRequest` to include `sentry-ipc:` and `sentry-ipc://start` / `sentry-ipc://scope`
2. **Disable Sentry in MAS entirely** — If Sentry can't work properly in the sandbox, gate renderer initialization behind `!isMasBuild()` to eliminate the spam
3. **Fix init timing** — Ensure `initSentry()` runs synchronously before `app.whenReady()` resolves, per the Sentry Electron docs

## Impact

Console noise only — Sentry is opt-in and disabled by default. No user-facing functionality is broken. But the spam makes debugging other issues harder during TestFlight testing.

Observed in: v1.0.0 build 14 (TestFlight)

mrangelmarino:
**MAS-refresh audit (no status change).**
- **State:** CSP blocks the `sentry-ipc` protocol in the MAS build, so renderer crash reports don't reach Sentry on MAS.
- **Blocker:** needs a CSP adjustment that's safe under the MAS sandbox.
- **Recommendation: land.** MAS is now the *free taste* client, and crash reporting on it is still worth having; verify the CSP fix is sandbox-compliant. Low-ish effort, real value.

Staying On Hold pending Angel's review of this audit.

PR #749, opened 2026-06-13
fix(sentry): use Classic IPC for runtime main-process init (#655)

### LINKED ISSUE #655: fix(sentry): runtime error-reporting toggle inits main SDK after app 'ready'
Follow-up from #391 (CSP `sentry-ipc:` fix).

## Problem

When a user enables error reporting **at runtime** (Settings → toggle on, while the app is running), the renderer fires `sentry:setEnabled(true)` → main `setSentryEnabled(true)` → `initSentry(true)` → `Sentry.init()`. Because `@sentry/electron/main` expects `Sentry.init()` **before** the `app` `ready` event, a late init logs:

> Sentry SDK should be initialized before the Electron app 'ready' event is fired

The launch-time path is already correct (`src/main/index.ts:28` inits before `whenReady`), and the renderer side has no such constraint — so for users who enable reporting and **restart**, crash reporting works fully. The warning only affects the same-session runtime-enable case, and is benign (no functionality broken — confirmed in #391).

## Constraint

The privacy invariant (CLAUDE.md) is **"Sentry never initializes unless `errorTracking.enabled === true`."** So the obvious "just init early always" fix is off the table — we can't initialize the SDK before the user opts in.

## Options to weigh

1. **Accept + suppress** the benign late-init warning (it's cosmetic for our JS-error use case).
2. **Restart hint** — when enabled at runtime, surface a subtle "takes effect on next launch" note; rely on the correct pre-`ready` path thereafter.
3. **Verify** whether late main-init actually degrades renderer→main envelope forwarding in the same session, and only then decide if more is warranted.

Low priority — console/UX polish on the free MAS client's opt-in crash reporting. Discovered while scoping #391.

mrangelmarino:
> ⚠️ **Privilege boundary: this PR touches the Electron MAIN process** (`src/main/sentry.ts`). The only change selects the `@sentry/electron` IPC mode at init time. **No sandbox/security settings are changed** — `contextIsolation`, `nodeIntegration`, CSP, and the privileged `sentry-ipc:` scheme registration are all untouched. The privacy invariant ("Sentry never initializes unless `errorTracking.enabled === true`") is preserved: this only changes *how* the SDK wires IPC when it does init, never *whether* it inits.

Closes #655

## What
When error reporting is enabled **at runtime** (Settings toggle while the app is already running), the renderer fires `sentry:setEnabled(true)` → main `setSentryEnabled(true)` → `initSentry(true)` → `Sentry.init()`. Because this happens *after* the Electron `app` `ready` event, `@sentry/electron`'s default IPC mode (`Both` = Protocol + Classic) tries to register the privileged `sentry-ipc:` scheme, which it requires to happen *before* `ready`, producing:

```
Sentry SDK should be initialized before the Electron app 'ready' event is fired
```

This change makes `initSentry()` select the IPC mode based on app state:

- **Pre-`ready` (launch path, `src/main/index.ts`)** — unchanged: default `IPCMode.Both`.
- **Post-`ready` (runtime opt-in path)** — `IPCMode.Classic`, which needs no pre-`ready` scheme registration and therefore does not emit the warning.

## Why it's safe
Renderer→main envelope forwarding is unaffected: the renderer prefers the preload-injected Classic IPC (`window.__SENTRY_IPC__`) and only falls back to the `sentry-ipc:` custom protocol when Classic isn't available. So crash reports still reach the main process in the same session, and the launch-time Protocol path (covered by #391's CSP fix) is unchanged.

## Files
- `src/main/sentry.ts` — choose `ipcMode` based on `app.isReady()`; add `IPCMode` import.

## Validation
- `electron-vite build` compiles cleanly (main/preload/renderer). Note: the full `npm run build` script also runs `build:skill`, which fails in this sandbox only because `zip` is not installed — unrelated to this change.

## Manual QA
1. Build and launch the app with error reporting **off** (`errorTracking.enabled` false / fresh profile).
2. Open Settings → General → **Error Reporting** and toggle it **on** while the app is running.
3. Open the main-process console/logs and confirm the `Sentry SDK should be initialized before the Electron app 'ready' event is fired` warning **no longer appears**, and `[Sentry] Main process initialized` is logged.
4. Trigger a main-process or renderer error (e.g. via a dev error path) and confirm a crash report is still captured in the same session (Classic IPC path works).
5. Fully quit and relaunch with the setting **on**; confirm startup is clean (pre-`ready` Protocol path) and `[Sentry] Main process initialized` logs without warning.
6. Toggle Error Reporting **off** and confirm `[Sentry] Main process disabled` logs and no further reports are sent.

_Conversation: https://app.warp.dev/conversation/f297d358-8879-48f9-aa89-ac6ad86eb6ce_
_Run: https://oz.warp.dev/runs/019ec2f8-411e-7306-88ac-d70afef0b984_

_This PR was generated with [Oz](https://warp.dev/oz)._

### DIFF
--- a/src/main/sentry.ts
+++ b/src/main/sentry.ts
@@ -1,4 +1,5 @@
 import * as Sentry from '@sentry/electron/main'
+import { IPCMode } from '@sentry/electron/main'
 import { app } from 'electron'
 import { is } from '@electron-toolkit/utils'
 
@@ -40,12 +41,24 @@ function beforeSend(event: Sentry.ErrorEvent): Sentry.ErrorEvent | null {
 export function initSentry(enabled: boolean): void {
   if (!enabled || initialized) return
 
+  // The default IPC mode (Protocol + Classic) registers a privileged `sentry-ipc:`
+  // scheme, which @sentry/electron requires to happen BEFORE the app 'ready' event
+  // and otherwise throws "Sentry SDK should be initialized before the Electron app
+  // 'ready' event is fired". The launch-time path (src/main/index.ts) inits before
+  // whenReady so the default is fine there, but the runtime opt-in path (Settings
+  // toggle → sentry:setEnabled) runs after 'ready'. In that case fall back to Classic
+  // IPC, which needs no pre-ready scheme registration. Renderer→main envelope
+  // forwarding is unaffected: the renderer prefers the preload-injected Classic IPC
+  // (window.__SENTRY_IPC__) and only uses the sentry-ipc: protocol as a fallback.
+  const ipcMode = app.isReady() ? IPCMode.Classic : IPCMode.Both
+
   Sentry.init({
     dsn: SENTRY_DSN,
     environment: is.dev ? 'development' : 'production',
     release: app.getVersion(),
     sampleRate: 0.5,
     maxBreadcrumbs: 30,
+    ipcMode,
     beforeSend
   })
   initialized = true


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/801ff5ce-27fa-4345-a323-44356fae44c5?refs=fe623077bd1f6ea0dc63a3b46bda436887371f32&source=prChecksComment) |

mrangelmarino:
/review 

PR #791, opened 2026-06-17
Release v1.6.5 — explorer, menus, theme, thinking, provenance, comments

### LINKED ISSUE #570: AI provenance for content not authored via insert/edit (create-from-scratch, MCP creates, paste)
## Motivation

The AI-authorship annotation system (unified in #548) records provenance — model, timestamp, optional rationale — for spans written via `insert` / `edit` / `suggest_edit`. But content that enters the document through other paths carries **no provenance trail**:

- **Create-from-scratch** (`create_and_open_file`) — the agent authors an entire new document, yet none of it is annotated. Caught during #467 Stage 6 QA.
- **MCP creates via Claude Desktop** — `create_and_open_file` / `insert` / `suggest_edit` invoked through the MCP bridge should record the same authorship trail as in-app calls.
- **Copy/paste** — pasted AI-generated text (e.g. from another Claude surface) bypasses the annotation system entirely.

## Proposed shape

1. Annotate the full body of documents produced by `create_and_open_file` as AI-authored (model + timestamp; the create intent can seed a rationale).
2. Route MCP-originated creates/edits/inserts (Claude Desktop → bridge) through the same provenance path as in-app tool calls.
3. On paste of a sizable block, offer a quick modal: "Mark this as AI-authored?" — attaches provenance when confirmed.

## Files (leads)

- `src/renderer/lib/tools/executors/editor.ts` — create executor
- `src/main/mcp/` bridge — MCP-originated tool calls
- `src/renderer/extensions/ai-annotations/` — annotation creation
- editor paste handling (`src/renderer/components/editor/` + tiptap paste rules)

## Acceptance criteria

1. `create_and_open_file` output is annotated as AI-authored (model + timestamp).
2. MCP creates/edits from Claude Desktop record provenance identically to in-app calls.
3. Pasting a block surfaces an optional "mark as AI-authored" modal that attaches provenance when confirmed.
4. Existing in-app insert/edit/suggest_edit provenance unchanged.

## Notes

Follow-up from #467 Stage 6 QA; relates to #548. Larger than the v1.2 agent-persona pass — milestone TBD. The copy/paste modal is the most open-ended part and may split into its own issue.

Refs #467 #548

### LINKED ISSUE #699: feat(comments): comment threading — replies, real resolved state, AI participation
## Summary

Comments become **threads**: users can reply to comments, and "Process" lets the AI participate like a real human collaborator — replying when a reply makes sense, editing when an edit makes sense, and resolving only when the comment is actually addressed.

## Current state

- Comments are TipTap marks with `{id, comment, createdAt}` (`src/renderer/extensions/comments/types.ts`) — no author, no replies, no resolved state. "Resolve" = mark deletion.
- Persistence: IndexedDB `comments` store (v8), keyed by documentId; comments never serialize into the `.md`.
- **Process** (`CommentPopover.tsx:155–160` → `useChat.ts:751–763`) builds a prompt, fires the normal chat loop, then **immediately deletes the comment mark** — the AI never replies, and resolution isn't an AI decision.

## Scope

1. **Data model**: `CommentData` gains `replies: {id, author: 'user' | 'ai', text, createdAt}[]` and a real `resolved: boolean`. Resolved threads persist (collapsed) instead of being deleted. Same IndexedDB store, richer values — no new object store needed.
2. **UI**: `CommentPopover` becomes a thread view — reply list, reply input, resolve control, collapsed state for resolved threads.
3. **AI participation**: new `reply_to_comment` tool in the registry (`src/shared/tools/registry.ts`, Editor-mode gated). Process stops host-side auto-deletion; the AI reads the comment and decides — reply, edit (`suggest_edit`), or both — and calls `resolve_comment` when done. `resolve_comment` sets `resolved: true` rather than unsetting the mark.
4. **Activity panel**: comment resolutions/replies become an activity type in the Activity tab (right sidebar), with a filter affordance like the superseded-toggle on annotations (#685 patterns).

## Folds in #665

Comment restoration currently fails when `markedText` spans block boundaries (`textBetween` separator vs `textContent` mismatch) and anchors by occurrence index (#648). This rework owns the restoration path, so the block-boundary matching fix lands here. Closing #665 in favor of this issue.

### LINKED ISSUE #700: feat(chat): extended thinking blocks in chat + dynamic streaming verbs
## Summary

Thinking blocks don't show up in the Chat panel — because extended thinking is **never requested**. Fix that end-to-end (enable → plumb → render), and fold in #556's dynamic verbing so the chat streaming surface gets one coherent upgrade.

## Root cause

- The stream call (`src/main/ipc.ts:930–936`) passes only `model`/`max_tokens`/`system`/`messages`/`tools` — no `thinking` param.
- The `contentBlockStart` handler only acts on `tool_use`; thinking blocks are silently ignored.
- No thinking event exists on the IPC bus (`llm:stream:chunk` / `tool-call*` / `complete` / `error` only) and `useChat.ts` has no handler.

So this is a feature never enabled, not a rendering bug.

## Scope

1. **Enable thinking** — `thinking: {type: "adaptive"}` on supporting models (Claude 4.6+; adaptive auto-enables interleaved thinking for tool loops — no beta header). On Opus 4.7/4.8, thinking text is omitted by default — pass `display: "summarized"` to get renderable content. Capability-gate per model (`src/shared/llm/models.ts` flags or the Models API capability tree); legacy models (if any remain offered) use `{type: "enabled", budget_tokens}` with `budget_tokens < max_tokens`. Raise the default `max_tokens` (currently 4096 in `ipc.ts`) to give thinking + output room.
2. **Preserve thinking blocks across the tool loop** — continuation requests must append the assistant's full content including thinking blocks unmodified (API requirement for tool-use turns with thinking). Audit the tool-loop continuation path in `useChat.ts` / main process.
3. **Plumb** — new IPC events (`llm:stream:thinking:start` / `:delta`; handle `signature_delta` internally) → preload → `useChat` → `chatStore`.
4. **Render** — collapsible thinking block in the chat transcript, collapsed by default with a subtle streaming indicator while thinking is in progress.
5. **Dynamic verbing (folds #556)** — Claude-Code-style contextual/dynamic verb text during tool calls and text generation (Thinking… / Drafting… / Searching… / Editing…), plus fluid text streaming. Existing primitives to build on: "Thinking..." pre-stream block (`ChatPanel.tsx:426–437`), "Drafting <tool>…" during tool-arg generation (`ChatMessage.tsx:15,50`), "<tool>..." while executing.
6. **UX** — thinking on by default; toggle lives alongside the model selector; new keyboard shortcut to flip it (proposal: ⌘⌥T — ⌘⇧T is taken by Reopen Closed Tab; verify against `menu.ts` and renderer handlers).

Supersedes #556 (closing it in favor of this issue).

### LINKED ISSUE #701: feat(ui): customizable '…' menus — hide, re-add, reorder with wiggle edit mode
## Summary

Make the "…" menus user-configurable: hide items, re-add them, and reorder them — with items wiggling slightly while in edit mode (iOS-style).

## Scope — the three kebab menus

| Menu | Location | Notes |
|---|---|---|
| Toolbar **More options** | `Toolbar.tsx:554–618` | The main target — New/Open/Save/Export/Settings/support links |
| FileListPanel **header overflow** | `FileListPanel.tsx:1112–1137` | Responsive view-toggle overflow — customization (which view toggles show + order) must compose with the existing overflow logic |
| ChatPanel **History** dropdown | `ChatPanel.tsx:271–324` | Static items only (e.g. New Chat); the dynamic conversation list is not customizable |

Right-click `ContextMenu`s (file rows, projects, favorites) are **out of scope**.

## Edit → wiggle → save flow (agreed in refinement)

1. A subtle **"Customize…"** item pinned at the bottom of each menu.
2. Selecting it keeps the menu open in **edit mode**: items wiggle (±1° CSS keyframe), each row gains a drag handle (reorder) and an eye/hide toggle.
3. Hidden items sink to a greyed **"Hidden"** section at the bottom, where they can be re-added.
4. Click anywhere outside (or a "Done" row) saves — no separate settings UI.

## Persistence

`settings.json` under `menuCustomization[menuId] = { order: string[], hidden: string[] }`. Each menu defines a registry of stable item ids; items not present in the saved order (new features) append visible at the end, so upgrades never hide new menu items.

### LINKED ISSUE #703: feat(files): explorer interactivity — New Folder, folder rename, greyed-out filetypes, dotfile toggle
## Summary

Make the file explorer panel more interactive: full "filesy" right-click on empty space, a New Folder action, folder renaming, and visibility of non-markdown files (greyed out) with a Finder-style dotfile toggle.

### 1. New Folder
No folder-creation path exists today — no `file:createDirectory` IPC at all. Add one (with `validatePath()`, MAS-sandbox-safe), exposed in the empty-space context menu and folder-row context menu.

### 2. Empty-space context menu: full file operations
Empty-space right-click already exists (`FileListPanel.tsx:1383–1494` wraps the scroll area; current items: New File / Paste / Add as Project / Add to Favorites). Expand it into a proper "filesy" surface: New File, **New Folder**, Paste, Show in Finder (current root), plus the existing project/favorite actions.

### 3. Folder rename
Files have inline rename; folder rows don't (`FileTree.tsx:480–519` has no Rename item). Add Rename + the inline-rename affordance to folder rows. Watch the knock-ons: open tabs, file watchers, and favorites/projects pointing at paths under the renamed folder.

### 4. Non-markdown filetypes, greyed out — always visible
`file:listDirectory` filters to `.md/.markdown/.txt` (`src/main/ipc.ts:515`). Decision from refinement: other extensions are **always visible but greyed out** (not behind a toggle). Greyed rows are not openable in the editor; their context menu offers **Show in Finder** and **Open Externally** (`shell.openPath`). Mind listing perf and the `hasChildren` peek (`ipc.ts:500–503`) when including all files.

### 5. Dotfiles — `⌘⇧.` toggle
Dotfiles are unconditionally skipped (`ipc.ts:472`). Decision from refinement: `⌘⇧.` toggles dotfile visibility, matching the Finder convention (revealed dotfiles render greyed like other non-md files; a dotfile `.md` renders normally). No existing binding conflicts found in `menu.ts` or renderer keydown handlers. Decide at implementation whether the toggle persists (setting) or is session-only.

### LINKED ISSUE #722: feat(appearance): per-mode theme selection (separate Light/Dark themes)
## Summary

Let users choose a **different theme for Light mode vs. Dark mode**. The Mode selector (Light / Dark / System) stays as the active-appearance control; the *theme* becomes per-mode. When **System** is selected, the per-mode theme assignments persist ("stick") and the app applies the Light-mode theme when the OS is light and the Dark-mode theme when the OS is dark.

New-install defaults change to:
- **Light → Mono**
- **Dark → Prose**
- **Mode → System** (unchanged)

## What already exists

Theming is already a structured `Appearance` object — this is an extension, not a rebuild.

- **Settings shape** (`src/renderer/types/index.ts:44-50`): `Appearance { color: ColorTheme; mode: ThemeMode; icon; migrationToastShown }`, where `ColorTheme = 'prose' | 'mono' | 'termy'` and `ThemeMode = 'light' | 'dark' | 'system'`. A single `color` currently applies to both light and dark.
- **System mode already works**: `resolveEffectiveMode` + `setupSystemModeListener` (`src/renderer/stores/settingsStore.ts:34-39, 186-206`) listen to `matchMedia('(prefers-color-scheme: dark)')` and re-apply on OS change. We reuse this listener as-is; only *which color it applies* changes.
- **All three themes ship full light + dark CSS** (`src/renderer/index.css`): `.theme-prose` / `.theme-prose.dark`, `.theme-termy` / `.theme-termy.dark`, Mono via `:root` / `.dark`. No theme is light-only or dark-only, so any theme can be assigned to either mode.
- **Theme application** (`applyAppearance`, `settingsStore.ts:43-55`): toggles `theme-{color}` + `dark` classes on `<html>`.
- **Settings UI**: `src/renderer/components/settings/AppearancePane.tsx`, under the **"Appearance"** tab (`SettingsDialog.tsx:254`). Sections: **Mode** (segmented Light/Dark/System), **Color** (3 ThemeCards), **App icon**.
- **Defaults** live in three spots that all must change: `settingsStore.ts:15-22` (`FRESH_INSTALL_APPEARANCE`), `src/main/ipc.ts:149` (main-process default), and `AppearancePane.tsx:24-25` (`DEFAULT_COLOR` / `DEFAULT_MODE`, used by "Reset to default").

### Data model
Replace the single `color` with per-mode fields, e.g.:
```ts
interface Appearance {
  lightColor: ColorTheme   // theme used when effective mode is light
  darkColor: ColorTheme    // theme used when effective mode is dark
  mode: ThemeMode          // 'light' | 'dark' | 'system'  (the active/default setting)
  icon: IconId
  migrationToastShown: boolean
}
```
`applyAppearance` picks `lightColor` or `darkColor` based on `resolveEffectiveMode(mode)`. The existing system-mode listener already re-fires on OS light/dark change, so per-mode themes "stick" under System for free.

### UI (Appearance pane)
The interaction the requester described: **select a mode, then select the theme for it.** Two layout options — decide during implementation:

- **(A) Two always-visible theme pickers** — a "Light theme" picker and a "Dark theme" picker, both shown regardless of the active mode. Simplest, and makes the System behavior obvious (you see both assignments at once). Mirrors VS Code's "Preferred Light/Dark Color Theme" model.
- **(B) Mode-scoped picker** — the theme cards apply to whichever mode you're currently editing; requires an explicit light/dark sub-toggle when Mode = System so both can be set.

Recommend **(A)** for clarity; **(B)** is closer to the literal "pick a mode, then pick its theme" phrasing. This is the one genuine open design decision. `ThemeCard` already previews in a given mode, so each picker can render its cards in the correct light/dark palette.

### Defaults (new installs)
`lightColor: 'mono'`, `darkColor: 'prose'`, `mode: 'system'`. Update all three default sites above and the "Reset all to default" comparison.

### Migration
Add a per-mode migration to `migrateOnDiskSettings()` (`settingsStore.ts:136-179`, same pattern as the existing v1.1→v1.2 `LEGACY_THEME_TO_APPEARANCE` map). For existing users, derive `lightColor = darkColor = <old color>` so **nothing visibly changes for them** — the new split defaults apply to fresh installs only. (Don't silently flip an existing Mono user's dark mode to Prose.) Settings are plain JSON, so no IndexedDB `DB_VERSION` bump is involved.

## Naming

The settings tab is **already named "Appearance"** — the macOS-standard umbrella term for exactly this (Light / Dark / Auto + theme). Recommend **keeping "Appearance"** rather than renaming the tab. The per-mode change mainly reshapes the inner **"Color"** section into a per-mode **"Theme"** picker (with "Light" / "Dark" sub-labels). If a more explicit combined label is desired, "Color Mode & Theme" works, but "Appearance" is the cleaner, more conventional choice.

## Acceptance criteria
- [ ] Light and Dark modes can each be assigned a distinct theme
- [ ] With Mode = System, both per-mode themes persist and the correct one applies live as the OS switches light/dark (no restart)
- [ ] New installs default to Light→Mono, Dark→Prose, Mode→System
- [ ] Existing users keep their current theme in both modes after upgrade (migration, no surprise change)
- [ ] "Reset all to default" reflects the new per-mode defaults

## Related
- Settings redesign epic #698 (this touches the Appearance pane — coordinate if the redesign lands first)

### LINKED ISSUE #723: fix(files): File Explorer bugs + comprehensive QA pass
## Summary

Three concrete File Explorer bugs/gaps, plus a follow-on testing initiative to broadly QA the explorer (mouse, right-click, keyboard, drag/drop, rename, multi-select, file actions). Related but distinct from #703 (explorer interactivity — New Folder / folder rename / greyed-out filetypes).

### 1. Pressing Enter to commit a rename scrolls the File Explorer to the top
- **Repro:** Scroll down in the file list, right-click a file → Rename, type a new name, press Enter.
- **Expected:** List stays scrolled to the renamed file (which should remain focused/visible).
- **Actual:** List jumps to the top.
- **Likely culprit:** `src/renderer/components/files/FileTree.tsx` — the rename-start effect calls `input.closest('[data-radix-scroll-area-viewport]')?.scrollTo(0, 0)` (line ~281) to undo `select()`-induced scroll. After commit, the list re-fetches/re-sorts and focus returns to the button via `focus({ preventScroll: true })` (~line 261), but the scroll reset and/or re-mount appears to lose the prior scroll position. Worth instrumenting both start and commit paths to confirm which one is the trigger the user is seeing.

### 2. No multi-select in the File Explorer
- Today you can only act on one file at a time. Add range-select (Shift+Click), toggle-select (Cmd/Ctrl+Click), and Select All (Cmd/Ctrl+A while focus is in the explorer).
- Context menu and keyboard actions (Delete, Move to Trash, Duplicate, Cut/Copy/Paste, drag) should operate on the selected set.
- Decide selection clearing rules (click empty area, switch root, focus loss).

### 3. After completing a file action, focus does not return to the File Explorer
- **Repro:** Right-click a file → Delete / Duplicate / Move to Trash / etc. After the dialog or action finishes, keyboard focus is lost (no obvious focused row in the explorer).
- **Expected:** Focus returns to the explorer, ideally on a sensible neighbor (next sibling after delete, the duplicated file after duplicate, the renamed file after rename).
- This matters for keyboard users — after any action you should be able to immediately arrow-key to the next file without a manual click.

## Testing follow-up

The explorer has grown organically (right-click menus, inline rename, drag/drop, keyboard nav, preview vs. permanent tabs, projects/favorites, MAS sandbox paths, watcher updates) and we don't have broad coverage. Two parallel pushes:

**Manual QA pass** (one focused session, ideally before next release):
- Single-click (preview tab) vs. double-click (permanent tab) promotion
- Right-click on file, folder, empty space, multi-selection
- Keyboard: arrow keys, Enter (open/rename), F2 (rename), Delete/Backspace, Esc, Tab, Cmd+A, Cmd+C/V/X
- Rename: start, commit (Enter), cancel (Esc), blur, invalid names, duplicate names, extension preservation
- Delete / Move to Trash / Duplicate / Show in Finder / Open Externally
- Drag/drop within the tree (if supported) and from Finder into the explorer
- Watcher behavior: external rename/delete/create should reflect in the panel
- Behavior across project roots, favorites, recent files
- Edge cases: deeply nested folders, long filenames, non-ASCII names, very large directories
- MAS build: confirm bookmark-based access still works for actions

**Automated coverage** (Playwright e2e under `e2e/`):
- At minimum: rename happy path + Enter-doesn't-scroll regression, multi-select happy path, focus-returns-after-action regression for delete/duplicate.

## Acceptance

- [ ] Bug 1: Rename → Enter preserves scroll position. e2e regression test added.
- [ ] Bug 2: Multi-select (Shift+Click, Cmd/Ctrl+Click, Cmd+A) works; context menu and keyboard actions operate on the set. e2e test for the happy path.
- [ ] Bug 3: After delete/duplicate/rename/move-to-trash, focus returns to a sensible row in the explorer. e2e regression test for at least delete.
- [ ] Manual QA checklist above run, with any new bugs filed as separate issues.

## Related

- #703 — feat(files): explorer interactivity (new folder, folder rename, greyed-out filetypes, dotfile toggle)
- #717 — feat(tabs): right-click menu (Open in File Explorer)
- #605 — feat(agent): File Explorer control tool

### LINKED ISSUE #727: Saving a document whose folder no longer exists throws uncaught ENOENT and silently fails (PROSE-K)
## Summary

Saving a document whose parent directory no longer exists throws an **uncaught `ENOENT`** that propagates to Sentry, and the save **fails silently** — no notification, no recovery prompt. Because autosave also fails silently and never clears the dirty flag, this is a **data-loss risk**, not just a crash log.

**Sentry:** [PROSE-K](https://soloist.sentry.io/issues/7543215870/) — `Error invoking remote method 'file:save': Error: ENOENT: no such file or directory, open '/Users/[redacted]'` (priority: high, unresolved, new).

## Real-world repro (reporter's narrative)

1. A `.md` blog post was open in a tab at `old-dir/post.md`.
2. An external `git checkout` brought in the latest, which **moved the file to a different directory** and removed `old-dir/` — so `old-dir/post.md` became a dead reference into a now-nonexistent directory.
3. The doc was reopened from the **Recents** list (which still pointed at the old, dead path).
4. A save (manual ⌘S, or autosave) targeted the stale path → `writeFile` → `ENOENT` (writing into a missing directory fails). The error surfaced to Sentry; nothing was shown in-app.

## Root cause

The error is on the **save** path (`file:save`), and nothing along it handles a missing parent directory:

**(a) Main handler throws raw** — `src/main/ipc.ts:197-201`:
```ts
ipcMain.handle('file:save', async (_event, path, content) => {
  const safePath = validatePath(path)        // only blocks `..` traversal
  await writeFile(safePath, content, 'utf-8') // no try/catch, no parent-dir check, no mkdir
})
```
If the parent directory is gone, `writeFile` throws `ENOENT` and the rejection crosses IPC unguarded.

**(b) Renderer manual-save doesn't catch** — `src/renderer/hooks/useEditor.ts:189-234` (`saveFile()`):
```ts
await window.api.saveFile(document.path, content) // line ~194, no try/catch
setDirty(false)
```
Call sites invoke it bare (no `await`, no `.catch`): `Editor.tsx:611` (⌘S) and `App.tsx:706` (menu Save). So the IPC rejection becomes an **unhandled promise rejection** → captured by Sentry's renderer integration → matches the `Error invoking remote method 'file:save'` format. **No `notificationStore` toast is shown** — the user gets no feedback the save failed.

**(c) Autosave fails silently and retries forever** — `src/renderer/hooks/useAutosave.ts:26-54` (`performSave`) *does* wrap the call in try/catch, but only `console.error`s (line ~49). Consequences:
- The ENOENT is swallowed (so autosave isn't the Sentry source — manual save is).
- `setDirty(false)` is never reached → the tab stays dirty → autosave keeps re-firing to the dead path every ~1.5s indefinitely.
- The document is **never persisted**; if the user later closes the tab believing autosave handled it, edits are lost.

**(d) Stale paths are never validated or pruned:**
- Session restore (`recoverSessionState`, `src/renderer/components/layout/App.tsx:239-306`; `tabStore` restore) recreates tabs with whatever path was persisted, including dead ones — no existence check.
- The **in-app Recents panel** (`src/renderer/components/files/FileListPanel.tsx:1177-1213`) renders the raw `settings.recentFiles` array and makes every entry clickable, **including dead paths**. The native macOS recents menu already filters these via `existsSync` in `src/main/recentFiles.ts:17`, but that filter isn't applied renderer-side.
- No `validatePath`/`existsSync`/`mkdir` reconciliation detects that a file/dir was moved externally.

## Proposed fix

Handle "the target path is gone" as a first-class, recoverable condition rather than an uncaught throw:

1. **Main `file:save` (`ipc.ts:197`)** — wrap in try/catch and return a **structured error** (e.g. `{ ok: false, code: 'ENOENT' }`) instead of throwing a raw rejection. Do **not** blindly `mkdir -p` the parent — the directory was deliberately removed by the checkout, so silently resurrecting it would create a ghost folder. Let the renderer decide.
2. **Renderer `saveFile()` (`useEditor.ts:189`)** — add try/catch. On `ENOENT`/missing-dir: show a `notificationStore` toast ("This file's folder no longer exists") and trigger the **Save As** flow so the user can relocate the document (no data loss). Ensure no unhandled rejection escapes; await/`.catch` at the `Editor.tsx`/`App.tsx` call sites.
3. **Autosave (`useAutosave.ts`)** — on a missing-path failure, **stop retrying** that tab (don't hammer the dead path every 1.5s), surface a single non-spammy "couldn't autosave" indicator, and mark the tab as needing relocation. Never leave the user believing edits are saved when they aren't.
4. **Detached-path handling** — when restoring a session or opening from Recents, validate the path; if it's gone, mark the tab as **detached/needs-relocation** (treat like an unsaved doc that needs a target) instead of silently retaining a dead path.
5. **Recents hygiene** — filter dead paths out of the in-app Recents panel (reuse the `existsSync` pattern from `recentFiles.ts:17`), and on an open/save failure, prune that entry + toast.

## Acceptance criteria
- [ ] Saving to a path whose directory was removed shows a clear notification and offers Save As — no uncaught error in Sentry
- [ ] No data loss: the user is never left believing a save/autosave succeeded when it didn't
- [ ] Autosave stops looping on a dead path and surfaces the failure once
- [ ] Session restore / Recents do not silently retain or offer dead paths; dead recents entries are pruned
- [ ] Regression coverage for save-to-missing-directory (e2e in an isolated `PROSE_USER_DATA_DIR` profile)

## Notes / related
- This is fundamentally a **stale-path robustness** cluster (open + restore + save + recents) surfaced by an external file move; the Sentry hit is the manual-save leg.
- Separate find while triaging: **[PROSE-J](https://soloist.sentry.io/issues/7515106237/)** — `ReferenceError: onAddFavorite is not defined` in `FileTree.tsx` `FileTreeItem` (a distinct, current crash in the file-tree favorites handler). Not part of this issue; flagging for its own ticket.

### LINKED ISSUE #777: fix(files): onAddFavorite is not defined crash in FileTree FileTreeItem (PROSE-J)
## Summary

`ReferenceError: onAddFavorite is not defined` crashes the file-tree favorites handler in `FileTree.tsx` (`FileTreeItem`). Surfaced as **Sentry PROSE-J** and flagged while triaging the stale-path cluster (#727); filing separately since it's a distinct, current crash with its own root cause.

## Root cause (lead)

`src/renderer/components/files/FileTree.tsx` — the `FileTreeItem` favorites handler references `onAddFavorite` that isn't in scope (not declared/destructured from props, or not threaded down from `FileListPanel`). The handler throws on invoke → unhandled rejection → Sentry.

Related latent gap found in the same component during the #727 Explore pass: `FileTree.tsx` is passed an **`onFileMove`** prop (from `FileListPanel.tsx:1470`) that the component **doesn't declare** — same family of prop-wiring gaps. Worth auditing the full `FileTreeItem` prop surface (favorites/projects/move handlers) for other undeclared/undefined references while here.

## Proposed fix

1. Audit `FileTreeItem`'s prop interface vs. what `FileListPanel` passes; declare/thread `onAddFavorite` (and verify `onFileMove`, `onRemoveFavorite`, project handlers) so every invoked callback is defined.
2. Defensive: guard optional callbacks (`onAddFavorite?.()`) so a missing handler degrades gracefully instead of throwing.
3. Regression: e2e covering the favorites context-menu action on a file row in an isolated `PROSE_USER_DATA_DIR` profile.

## Acceptance

- [ ] Add-to-Favorites from the file-tree row no longer throws `onAddFavorite is not defined`
- [ ] `FileTreeItem` prop surface audited — no undeclared/undefined callback references (incl. `onFileMove`)
- [ ] e2e regression for the favorites context-menu action

## Refs

- Surfaced during #727 triage (distinct from #727's stale-path ENOENT cluster)
- Sentry: PROSE-J

mrangelmarino:
## v1.6.5

Bundles nine feature/fix PRs from the parallelization run, integrated on
`release/v1.6.5` and verified end-to-end via a human-in-the-loop QA pass
(Clusters A/B/C + regression spot-checks) against both the dev build and the
packaged production build.

### What's in it

- **#723** — file explorer: rename keeps scroll + focus, multi-select, focus-return after ops
- **#703** — explorer interactivity: New Folder, folder rename, greyed non-md, `⌘⇧.` dotfile toggle
- **#727** — stale-path data-loss recovery (autosave stops looping, friendly Save As)
- **#777** — favorites crash from the chat inline FileTree
- **#701** — customizable menus, rebuilt as a unified drag-reorderable toolbar + file-explorer header
- **#722** — per-mode (Light/Dark) theme with live System switching + migration
- **#700** — extended thinking blocks (collapsible, model-gated toggle)
- **#570** — AI provenance for create + paste (annotations + "Mark as AI-authored?" modal)
- **#699** — comment threading redesign (replies, Process, resolved state, Activity entries)

Closes #723
Closes #703
Closes #727
Closes #777
Closes #701
Closes #722
Closes #700
Closes #570
Closes #699

### Defects caught & fixed during HITL QA (on this branch)

- `7c75b8f` — multi-select Cmd+Click toggle-off left the removed row as primary/anchor (#723)
- `1e94ea6` — folder rename now migrates open tabs / favorites / projects instead of orphaning them (#703)
- `8f00f89` — **two** bugs blocking "thinking off": the main process omitted the param (defaults on for adaptive models → now sends `{type:'disabled'}`), and the renderer send path read a stale `useCallback` closure (#700)

### Non-blocking follow-ups filed (pre-existing / deferred — do NOT gate this release)

- #797 — multi-select Duplicate action + residual root-switch selection
- #798 — footnote input rule leaves a stray `[` when `[^N]` abuts text (pre-existing dep behavior)
- #799 — footnote inline references lost on markdown round-trip (pre-existing parse gap)

### Notes

- #570 Leg 3 (MCP-originated create) couldn't be exercised against the isolated QA profile (MCP socket path); it rides Leg 1's annotation-on-create path, verified live.
- Version bumped to **1.6.5**; MAS `buildVersion` → **33** for the parallel TestFlight build. `electron-builder.yml` touched → expect the security-gate flag (human merge approval).


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/a17205a6-e2f0-4476-83a4-711979053112?refs=5083f8de8d13712ab9a11eec7f176d37853cea0e&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
/review 

Issue #403, opened 2026-04-05
feat(remarkable): support PDF and text document sync


mrangelmarino:
## Problem

reMarkable tablets support PDF uploads and text documents in addition to handwritten notebooks. The current sync pipeline only processes `.rm` stroke files via OCR, so PDFs and text documents sync successfully (download works) but display with empty content bodies — just the metadata header.

## Expected behavior

- **PDFs**: Extract text content or render pages. The `.rm` file in a PDF notebook may contain annotations, but the PDF itself has extractable text.
- **Text documents**: reMarkable now supports plain text documents. These should sync their text content directly without needing OCR.

## Current behavior

- Notebook downloads correctly (13KB+ of data)
- OCR service processes the page "successfully" (`pages=1, failed=0`)
- But returns empty markdown because the data isn't handwritten strokes
- File is written with metadata header only, no content body

## Discovery

Found during QA of PR #398 (#266 sync UX). User synced "Outline" (a text document in Writing > Beyond Left and Right) — downloaded fine but showed empty content.

## Implementation notes

- `fileType` is already available on each notebook from the cloud listing (`notebook`, `pdf`, `epub`)
- The sync pipeline in `src/main/remarkable/sync.ts` could branch on `fileType` before calling `processNotebookWithOCR`
- For PDFs: extract text via a PDF library, or convert pages to images for OCR
- For text/epub: extract text content directly from the document format

PR #807, opened 2026-07-01
feat(remarkable): sync typed-text documents (Part of #403)

mrangelmarino:
## What & why

reMarkable **typed-text documents** (created with the Type Folio keyboard) sync but display with empty bodies — the bug captured in #403. They're stored as `fileType: 'notebook'`, but their content lives as digital text inside reMarkable's v6 `.rm` binary scene (`RootTextBlock`), *not* as handwriting strokes. The OCR pipeline renders strokes, so it produces a blank page for typed text.

This PR extracts that typed text directly, so these documents finally sync with real content.

**Scope:** typed text only. PDF/EPUB extraction (the rest of #403) is deferred — this is **Part of #403**, not a close.

## How

- **`src/main/remarkable/rm-scene-parser.ts`** (new) — a zero-dependency v6 `.rm` parser, ported from the **MIT-licensed [`rmscene`](https://github.com/ricklupton/rmscene)** by Rick Lupton (attribution in the file header). It reads the `RootTextBlock`, decodes the CRDT text sequence (with topological ordering), and maps paragraph styles to markdown (heading/bold/bullet/checkbox). It **never throws** — any malformed/unexpected input returns a safe fallback so the caller routes the page to OCR instead.
  - *Not* derived from the GPL-3.0 TypeScript port in the TimDommett/Remarkable-Sync Obsidian plugin (Prose is MIT).
- **`processNotebookContent`** (renamed from `processNotebookWithOCR`) parses typed text from every page's bytes each sync and OCRs **only** handwriting pages. Typed text wins per page for mixed notebooks. Reading every page bypasses the OCR cache, which can hold stale blank results for typed docs synced before this feature.
- New `extraction` metadata field (`typed-text` | `ocr` | `mixed`) and a broadened `needsExtraction` gate (no longer requires OCR to be configured) that **re-processes legacy entries once**, healing already-synced typed docs with empty bodies.
- Renderer: `'extracting'` sync phase (spinner) + a type-aware tooltip ("typed text" vs "OCR").

**Design property — graceful degradation:** typed-text extraction is purely additive. If the parser ever fails (e.g. a future firmware format change), the page falls back to the existing OCR path. Worst case = today's behavior; never a crash or regression.

No new dependencies. No `client.ts` changes (the `.rm` files are already in the downloaded ZIP).

## Testing

- `npm run build` passes; dev build launches Electron cleanly.
- **9 new pure-Node Playwright tests** (`e2e/electron.remarkable-{parser,sync}.spec.ts`) run against **real reMarkable `.rm` fixtures** vendored from MIT rmscene (`e2e/fixtures/remarkable/`), cross-checked against rmscene's own expected extractions — covering plain/heading/bullet/bold styles, CRDT concurrent-author ordering, mixed typed+handwriting pages, the `typed-text`/`mixed` classification, the graceful no-OCR skip, and corrupt/empty input.

```
npx playwright test electron.remarkable-parser electron.remarkable-sync
```

## Manual verification

Sync a typed reMarkable doc (e.g. "Outline"): it should open with real text and be clickable — even with OCR disabled, since typed text needs no OCR. Already-synced empty typed docs heal on the next sync. Pure-handwriting notebooks still OCR exactly as before.


mrangelmarino:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/mrangelmarino/pr-checks/def72c8b-120b-4757-a4b5-e6f0fda9b15e?refs=df40fadb01734b37b176f1f65aa273517d8d7d28&source=prChecksComment) |

mrangelmarino:
/review 

mrangelmarino:
Added a second commit fixing reMarkable **read-only tab handling**, surfaced by this feature (typed docs are now openable for the first time, which exposed a pre-existing bug):

- Synced docs bypassed the tab system, so only one could open at a time and it never became a stable tab.
- Read-only was a floating global flag cleared on every tab navigation — so switching back to a reMarkable doc left the **hidden OCR source file editable in place**, where a Cmd+S would clobber sync-managed output.

Fix: read-only is now **derived from the path** (any `.remarkable/<id>/<name>.md` is inherently a read-only OCR view — new `lib/remarkablePath.ts`, unit-tested), and reMarkable docs open through `openFileInTab` so each gets its own tab. Verified live: multiple docs open in separate tabs, read-only survives tab switches, and normal file editing is unaffected.

mrangelmarino:
/review 

mrangelmarino:
Addressed the review feedback in `7608df1`:

- **Re-processing loop (the real finding):** a handwriting notebook skipped without OCR now records `extraction: 'ocr'` on the graceful skip, so `needsExtraction` keys off *having a recorded outcome* rather than *having content* — no more re-reading/re-parsing its `.rm` files every sync. A second gate clause re-triggers it if OCR is configured later (and a content change re-triggers via `needsDownload` regardless). Covered by a strengthened integration test.
- **v10+ version parse:** `readHeader` now parses the full version token, not a single digit.
- Left the `toposortItems` O(n²) note as-is (acknowledged; pages are small).

All reMarkable specs green (11 passing).

mrangelmarino:
/review 

mrangelmarino:
Two more fixes from live QA (both verified in the running app):

- **`6bc8785` — sync freshness:** `connect()` caches the rmapi client for the whole session and `listItems()` reused the root hash from launch, so tablet edits that reached the cloud *after* Prose started weren't detected until a restart. Now passes `refresh=true` to re-fetch the current root hash (content cache preserved, so no needless re-downloads).
- **`df40fad` — read-only across restore:** OCR read-only was applied only at tab-nav sites; session restore on launch loaded the active doc via a raw `useEditorStore.setState()` and came up **editable** (a stray save would clobber the hidden sync file). Consolidated read-only derivation into the store setters (`setDocument`/`hydrateFromDraft`/`setPath`) as the single source of truth, and routed restore through `setDocument`. Removed the now-redundant per-nav-site derivation.

mrangelmarino:
/review 

